PR #6, opened 2023-01-06
Add software infrastructure

rafmudaf:
This pull request adds the initial infrastructure for OpenTurbine. The following primary components are added:

- Online documentation
- CMake cross-platform build system
- Continuous integration through GitHub Actions
- Testing driven by CTest using the Google Test framework

Since OpenTurbine itself is not yet formulated as a software framework, I've created a simple 1D heat conduction solver to test all systems. The infrastructure described here will be expanded as OpenTurbine is further developed, but this should serve as the foundation for ongoing development.

## Documentation

The documentation is hosted through GitHub Pages and built and deployed automatically when any file in `docs/` is modified. It uses the Jupyter Book framework that is built on top of Sphinx. Jupyter Book in particular allows for writing documentation in markdown or restructured text. It has support for additional markdown features through MyST and it integrates with all Sphinx extensions plus additional extensions specific to this framework.

The config file is at `docs/_config.yml`. The section `sphinx:` allows for using any configuration parameters used directly in Sphinx. This should be familiar to the ExaWind team as most ExaWind projects also use Sphinx.

A basic structure for the documentation can be seen at https://exawind.github.io/openturbine/intro.html.

**Note**: I will be adding to the documentation over the coming week as this pulls request is finalized.

## CMake

The build system entirely uses CMake, and the project was adapted from AMR-Wind. Some key features are:

- Version generation based on git status
- Automatic setting of vendor-specific compiler flags
- Linking to dependencies (currently Kokkos and Google Test)
- Multitarget build configuration to support improved testing

OpenTurbine-specific CMake variables are generally prefixed with `oturb` and flags/settings are prefixed with `OTURB`. The currently included targets are given in the table below.

| Target         | Type | Description |
| ----------- | ----- | ----------- |
| openturbine  | Exe | Primary executable including the `main`    |
| openturbine_obj   | Lib | Library including all code used by `main`    |
| openturbine_unit_tests   | Exe |  Unit test driver executable including all tests and links to the Google Test   |

In addition to the typical CMake flags, the following flags are available for specific hardware: `OTURB_ENABLE_<TYPE>` for each of CUDA, DPCPP, OPENMP, ROCM. `OTURB_PRECISION` can be DOUBLE or SINGLE. `OTURB_ENABLE_ALL_WARNINGS` adds `-Wall` or equivalent compiler flag. `OTURB_ENABLE_TESTS` adds the testing infrastructure to the build.

## Continuous integration

Automated testing is configured through GitHub Actions at `.github/workflows/ci.yaml`. The first step in the workflow is to run a static linter on the whole project with `clang-format`. If any changes are required, this step and the entire CI run will fail. The following step builds the project and runs the full test suite. Kokkos is installed through Spack, and Spack is directly configured in the workflow.

To save time installing dependencies, the entire `~/.spack` directory is cached. If there's a cache miss (for example, if a dependency has changed, the cache exceed its stale shelf life of a week, or if the cache limit of 10GB is exceeded), Actions will reinstall Spack and Kokkos, and this will take about 20 minutes. The workflow must finish successfully in order for the cache to save, so it can be useful to remove all other steps in the workflow if the cache needs to be repopulated.

See the test results live at https://github.com/Exawind/openturbine/actions.

## Testing

The test suite uses the [Google Test](https://google.github.io/googletest/index.html) framework and is integrated with CTest. All unit test files are included in `tests/unit_tests/`. The test driver is `utest_main.cpp`, and this collects all the other tests and creates the support needed to filter, run, and report tests. A help menu can be seen with `./openturbine_unit_tests --help`. All tests are integrated into CTest through the listing at `tests/CMakeLists.txt`.

# TODO

- [x] Add basic compiling instructions and dependencies to the documentation including README
- [x] Add more details on expectations from developers, reviewers, and users
- [x] Create stubs in the documentation for API, input file, tutorials, etc


.clang-format
```diff
@@ -0,0 +1,123 @@
...
+  SplitEmptyRecord: false
+  SplitEmptyNamespace: true
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Custom
+BreakInheritanceList: BeforeComma
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializers: BeforeComma
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: true
+ColumnLimit:     80
```
ddement, line 50: Do we want this strict of a line length restriction?  Strict enforcement of an 80 column limit can lead to harder to read formatting.  I personally prefer setting this value to 0 and enforcing line limits during code review based on situational style preferences and readability.  If we do want to enforce a hard cap, I'd be in support of setting the column limit to a higher value (say 200) to allow some flexibility.

CMakeLists.txt
```diff
@@ -0,0 +1,169 @@
+
+############################ BASE ######################################
+
+cmake_minimum_required (VERSION 3.20 FATAL_ERROR)
+project(OpenTurbine CXX C)
```
jrood-nrel, line 5: I think we can remove `C` from here.

jrood-nrel:
Doesn't have to be in this PR, but I think we should add CodeQL as soon as possible. https://github.com/Exawind/amr-wind/pull/761

jrood-nrel (approved):

.clang-format
```diff
@@ -0,0 +1,123 @@
...
+SpaceBeforeParens: ControlStatements
+SpaceBeforeRangeBasedForLoopColon: true
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 1
+SpacesInAngles:  false
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard:        Cpp11
```
ddement, line 117: Should we set this to Cpp17? (I don't know what this option actually does in practical terms)

CMakeLists.txt
rafmudaf: Good catch, fixed.

cmake/GetGitRevisionDescription.cmake
```diff
@@ -0,0 +1,180 @@
+# This file was retrieved from https://github.com/rpavlik/cmake-modules/
```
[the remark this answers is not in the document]
rafmudaf: In this case, the `.in` is an input to a function in this file and the `configure_file` command is called directly, so I think both must be committed. This was borrowed from AMR-Wind (see [here](https://github.com/Exawind/amr-wind/tree/main/cmake)).

.clang-format
rafmudaf: This was taken directly from [AMR-Wind](https://github.com/Exawind/amr-wind/blob/main/.clang-format). I agree that 80 characters is pretty restrictive, but I'll defer these decisions to those who will be working in this code directly in the future. There are other decisions that impact this such as variable and namespace conventions. For example, I called the primary namespace `openturbine::` which itself is pretty long and takes up a large portion of an 80 column limit.

jrood-nrel: This is the same file as used in AMR-Wind, Nalu-Wind. and the exawind-driver. This is the only objection we've had to it so far. I don't really have a preference, but if this is an issue, we can address it later in all projects.

.clang-format
jrood-nrel: I agree, and I also don't know what it really does.

docs/dev_guide.md
```diff
@@ -0,0 +1,73 @@
+# Developer's Guide
+
+Developers of OpenTurbine should understand the workflows described here.
+Adherence to these processes ensures high quality code with minimal maintenance.
+
+## Version control
+
+OpenTurbine uses git with a basic version of git-flow. The `main` branch is
```
ddement, line 17: How will we determine when code should be merged into main?  If we ensure that all tests always pass and our test coverage is adequate, then a more stable main branch is possibly redundant.  To support official releases, we can use specific release branches, which will offer stability and possibly V&V qualification opportunities to people who need it.

rafmudaf: In concept, I'm ok with this approach. However, I think it relies on adhering to strict gate keeping. Pull requests must be thoroughly reviewed and well tested prior to merging. This means someone or a group of people will have to assume the role of enforcing the rules even in the face of deadlines and other work piling up. In the past, I've found that code is considered complete in order to meet a deadline (i.e. milestone) rather than when it is fully implemented and verified. The ultimate question may be is the team willing to miss a milestone to enforce code quality? If so, then merging to `main` sounds good to me and certainly simplifies the git workflow. If not, building in the redundancy of a `develop` branch can add a buffer to catch less than good code.

src/heat_solve.H
```diff
@@ -0,0 +1,29 @@
+#ifndef HEAT_SOLVE_H
+#define HEAT_SOLVE_H
+
+#include <vector>
+
+namespace openturbine::heat_solve {
```
ddement, line 6: I am assuming that this code will not be sticking around once we have OpenTurbine code.  If it will be kept long term, it needs significant work to bring it up to a modern, sustainable standard (for example, the use of raw pointers and malloc is unnecessary in this context).  Let me know if you need help in this.

tests/unit_tests/utest_main.cpp
```diff
@@ -0,0 +1,19 @@
...
+ *  Entry point for unit tests
+ */
+
+#include "gtest/gtest.h"
+#include "OpenTurbineTestEnv.H"
+
+//! Global instance of the environment (for access in tests)
+openturbine_tests::OTurbTestEnv* utest_env = nullptr;
+
+int main(int argc, char** argv)
```
ddement, line 11: We need to call Kokkos::initialize before we can run any unit tests using Kokkos constructs and finalize after all tests are run (you can only call each one once).  This is most easily done in the main function.

src/heat_solve.H
rafmudaf: Definitely not sticking around. It's a toy problem in order to demonstrate the unit test infrastructure.

docs/dev_guide.md
alanw0: There are valid reasons for having a main branch that developers don't push directly to, such as if we have nightly testing that covers more builds or runs verification problems that developers can't reasonably cover in pre-push testing.
The discussion about milestone vs good code is a larger discussion, maybe we can discuss more at the in-person mtg on the 25th. I would argue a couple of points: 1) let's not report a milestone using code that isn't fully implemented and verified. milestone calculations should be repeatable and correct. 2) I argue that good development practices such as TDD should make us go faster overall, not slower. So we shouldn't need to choose between meeting a milestone or writing good code. :-) Of course I realize there are devils in the real-world details and good intentions can melt down sometimes...

.clang-format
ddement: I know that we've had the 80 character limit in Nalu-Wind lead to some less ideal line breaks when working on it.  This is especially true when working with longer variable names and some of C++'s more verbose syntax.  In talking with some other STK guys, 120 or 150 seems like a more practical limit.  There's no need to come up with a definitive answer now - just something to keep in mind.

tests/unit_tests/utest_main.cpp
rafmudaf: Yes good catch, I've fixed this. The primary infrastructure is here for the unit testing, but this should be modified into a form that is more appropriate for the architecture of the software when that is decided.

PR #13, opened 2023-03-08
Set Kokkos_DIR environment variable based on installation location from Spack

### LINKED ISSUE #12: Kokkos dir statically set in CI
Currently, the Kokkos directory is statically set in the GitHub Actions configuration [here](https://github.com/Exawind/openturbine/blob/main/.github/workflows/ci.yaml#L60). However, when the installed version changes and the configured directory does not, then there's a cmake configuration error since it cannot find the Kokkos library.

A good fix would be to dynamically set the `Kokkos_DIR` environment variable based on the installation location from Spack. Is this information available after the spack install?

faisal-bhuiyan:
Fixes #12. 

Previously, the Kokkos directory path (`Kokkos_DIR`) was hardcoded into the Github Actions configuration file, which caused `CMake` configuration error when there was a mismatch. Here, the path is set dynamically from `Spack` to fix the issue.

Specifically: 
- the Kokkos installation directory is searched via the command `spack find -p kokkos`
- the installation path is subsequently extracted by piping commands via `grep -m 1 kokkos` (to filter the output to show only the lines containing the installation path of the Kokkos package)
- `awk '{print $2}'` (to extract the second column of the filtered output, which contains the installation path)

The Github Actions result shows the [CI workflow was successful](https://github.com/faisal-bhuiyan/openturbine/actions/runs/4368158392) with the above changes included into it.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,10 +1,10 @@
 name: OpenTurbine CI
 
-on:
-  push:
-    branches: [main develop]
-  pull_request:
-    branches: [main develop]
+on: push
+  # push:
+  #   branches: [main, develop]
+  # pull_request:
+  #   branches: [main, develop]
 
 jobs:
   Formatting:
@@ -53,6 +53,7 @@ jobs:
       run: |
         . ~/.spack/Spack/share/spack/setup-env.sh
         export Kokkos_DIR=$(spack find -p kokkos | grep -m1 kokkos | awk '{print $2}')
+        # echo $Kokkos_DIR
         # ls $Kokkos_DIR
         mkdir build
         cd build
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,4 +1,4 @@
-name: OpenTurbine CI
+name: OpenTurbine-CI
 
 on: push
   # push:
@@ -8,7 +8,7 @@ on: push
 
 jobs:
   Formatting:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
     - name: Clone
       uses: actions/checkout@v3
@@ -21,7 +21,7 @@ jobs:
         clangFormatVersion: 14
   CPU:
     needs: Formatting
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
     - name: Cache install Spack packages
       id: cache-install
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -49,15 +45,14 @@ jobs:
...
     - name: Clone
       uses: actions/checkout@v3
-      with: 
+      with:
         submodules: true
     - name: Test
       run: |
         . ~/.spack/Spack/share/spack/setup-env.sh
-        export Kokkos_DIR=/home/runner/.spack/Spack/opt/spack/linux-ubuntu22.04-x86_64_v4/gcc-12.1.0/kokkos-3.7.01-zhffzchmugzhigeekgnatfeikkrjwmfw
+        export Kokkos_DIR=$(spack find -p kokkos | grep -m1 kokkos | awk '{print $2}')
```
rafmudaf, line 55: @faisal-bhuiyan thanks, I think this is good! Please explain the sections of this command either in a comment in the yaml file or in the pull request description for the benefit of future developers.

faisal-bhuiyan: @rafmudaf Please find the requested explanation added to both places you mentioned. Let me know if there is anything else we need to do here.

.github/workflows/ci.yaml
```diff
@@ -49,15 +45,15 @@ jobs:
...
     - name: Clone
       uses: actions/checkout@v3
-      with: 
+      with:
         submodules: true
     - name: Test
       run: |
         . ~/.spack/Spack/share/spack/setup-env.sh
-        export Kokkos_DIR=/home/runner/.spack/Spack/opt/spack/linux-ubuntu22.04-x86_64_v4/gcc-12.1.0/kokkos-3.7.01-zhffzchmugzhigeekgnatfeikkrjwmfw
+        # Find the Kokkos installation directory via spack & extract the path via grep & awk
```
rafmudaf, line 55: Thanks @faisal-bhuiyan. This is fairly high-level, so can you elaborate on the details? For example, I think something like this would be helpful to future readers:

```
`spack find -p` lists the package version and installation path
`grep -m 1 kokkos` extracts the line with the first instance of "kokkos"
`awk ...` extracts the second element of the grep and puts that into the env variable
```

The reason I'm pushing for this is that, while this works well, it's using a result from Spack that is probably not intended to be used to set an environment variable. That is to say that the developers of Spack may not consider the output of `spack find` to be a stable API. If this were parsing some environment variable set by Spack for this purpose, then I would consider it within the scope of the Spack documentation to instruct the usage. Here, though, if the Spack output changes format, this line will likely break and the reason may not be obvious to other OpenTurbine developers.

rafmudaf: Ah disregard, I just saw that you added similar information in the pull request description. My apologies, and thanks for putting it there!

rafmudaf (approved):

PR #20, opened 2023-04-12
Add development plan documentation

rafmudaf:
# Development plan

This pull request adds the initial development plan to the documentation. The included sections are listed below, and a live version of the documentation can be seen on [my fork's docs](http://www.rafmudaf.com/OpenTurbine/dev_plan.html).

- Background and overview
- Development priorities and use cases
- Programming language and models
- Design drivers and considerations
- Guidelines for Software Quality
- Application Programming Interface (API)
- Key numerical algorithms
- Verification and validation cases
- Target baseline turbines
- High-level development timeline


docs/dev_plan.md
```diff
@@ -0,0 +1,307 @@
...
+simulation code designed to meet the research needs of WETO and the broader
+wind energy community for land-based and offshore wind turbines. OpenTurbine
+will provide high-fidelity, highly performant structural dynamics models that
+can couple with low-fidelity aerodynamic/hydrodynamic models like those in
+OpenFAST, and high-fidelity computational fluid dynamics (CFD) models like
+those in the WETO and Office of Science supported ExaWind code suite.
+OpenTurbine will be designed to address shortcomings of wind turbine structural
+models and codes that are so important to the success of WETO modeling efforts.
+
+## Development priorities and use cases
```
rafmudaf, line 28: I covered most of this and expanded on it below. @michaelasprague I think we should either remove this section in favor of "Design Drivers and Considerations" or at least incorporate it there.

docs/dev_plan.md
```diff
@@ -0,0 +1,307 @@
...
+memory efficient and will enable vectorization and parallelization at multiple
+levels. They will be data-oriented in order exploit methods for accelerated
+computing including high utilization of chip resources (e.g., single
+instruction multiple data [SIMD]), parallelization through GPU or other
+hardware, and support for memory-efficient architectures such as the ARM-based
+Apple M-series chips. 
+
+*Use-case priority:* Time-domain simulation of land-based wind turbine dynamics coupled to computational fluid dynamics for fluid-structure-interaction.
+
+## Programming language and models
```
rafmudaf, line 44: Move below the "Design" section

docs/dev_plan.md
```diff
@@ -0,0 +1,323 @@
...
+either by the CPU or a hardware accelerator. The graphic below
+illustrates the difference between structures of arrays and arrays
+of structures.
+
+```{image} ./static/AoS_SoA.pdf
+:alt: aos_soa
+:width: 400px
+:align: center
+```
+
```
faisal-bhuiyan, line 184: The above figure is not showing up for me. We might need to add the figure to the directory.

image

docs/dev_plan.md
```diff
@@ -0,0 +1,323 @@
...
+A key design consideration of `OpenTurbine` is computational efficiency
+or performance. Both the quantity of work and the efficiency of data should
+be considered and measured (i.e. profiled) during any development effort.
+A modular architecture should support offloading computationally expensive
+tasks to hardware accelerators or specialized libraries, and support
+multiple options for doing so depending on user configurations and
+the computational environment.
+
+Similar to modularity in the architecure, expensive tasks should be
+structured in a kernel form. This low-level design pattern combines
```
faisal-bhuiyan, line 154: Just wondering if it might be appropriate here to refer to an external link that describes the kernel form with example?

docs/dev_plan.md
faisal-bhuiyan: Thanks for fixing the figure. It looks like a great addition to the description here. Just a minor suggestion, would it add to the clarity of the figure if we added data types to the array elements such as done [here](https://stackoverflow.com/questions/8377667/is-the-memory-allocated-for-struct-members-continguous-what-if-a-struct-member/8377717#8377717)?

faisal-bhuiyan:
> The included sections are listed below, and a live version of the documentation can be seen on [my fork's docs](http://www.rafmudaf.com/OpenTurbine/dev_plan.html).

@rafmudaf Do you mind generating the documentation one more time before we merge this PR please? Thanks for your time!

docs/dev_plan.md
rafmudaf: Sure I'm happy to add these details. However, there's a nuance that's worth clarifying. I've understood the benefit from SoA to be that sections of an array can be loaded to the cache at a time, and the contiguity being that many adjacent elements of an array are used in a computation (thus leading to SIMD). In the example described in that SO post, they're suggesting that avoiding the padding between different data types is the benefit. I suppose it's a combination of both in the end.

rafmudaf: Ok I just read the SO post you linked to in the text (https://stackoverflow.com/questions/17924705/structure-of-arrays-vs-array-of-structures) and that clarifies it a bit more.

faisal-bhuiyan: Looks good! Thanks for taking the time to expand on the concept.

ashesh2512 (approved):

faisal-bhuiyan:
> @faisal-bhuiyan Can we please remove highlighting of OpenTurbine, OpenFAST, etc?

Yup, done in the above commit.

docs/dev_plan.md
```diff
@@ -2,6 +2,110 @@
...
+will be designed to couple the beam finite element models and point-mass
+elements (e.g., representation of the nacelle) to a CFD mesh.  We will leverage
+the mesh mapping that was implemented and tested in the ExaWind codes. The
+representation of the turbine geometry is handled within the fluid solver,
+either as a three-dimensional surface mesh for high-fidelity geometry-resolved
+simulations, or as an actuator-line for mid-fidelity simulations. Those
+algorithms will work well for problems where a floating platform is represented
+as a point-mass. For geometry-resolved floating-platform offshore simulations,
+the API will be expanded to handle mapping from the structure surface mesh (if
+solid/shell elements are used) to the CFD mesh.
```
alanw0, line 253: Why this caveat for solid/shell elements? Nalu-wind's current interaction with OpenFast maps to/from sides/faces of the surface.

docs/dev_plan.md
```diff
@@ -2,6 +2,110 @@
...
+rotate with the hub. These models together constitute a set of
+differential-algebraic equations (DAEs) in the time domain.  We will build on experiences gained with OpenFAST and the nonlinear-beam-dynamics module, BeamDyn.
+
+For time integration of the index-3 DAEs we will use the generalized-alpha
+algorithm now established in BeamDyn which allows user-controlled numerical
+damping.  See Arnold and Brüls (2007) for details.
+
+The primary model for turbine blades and tower, which are slender and flexible
+structures, will be nonlinear finite elements based on geometrically exact beam
+theory (GEBT), which includes include bend-twist coupling. Spatial
```
alanw0: typo: 'includes include'

docs/compiling.md
```diff
@@ -19,11 +19,13 @@ for linking and runtime path searches.
   - GCC 5 or higher
   - LLVM Clang 7 or higher
```
alanw0, line 20: These are badly outdated. Do we need to list specific versions?

docs/dev_plan.md
```diff
@@ -0,0 +1,339 @@
...
+
+#### Data-oriented design
+
+OpenTurbine developers designing new algorithms and data structures
+should become familiar with the concepts of [data-oriented design](https://en.wikipedia.org/wiki/Data-oriented_design),
+particularly [structures of arrays vs. arrays of structures](https://stackoverflow.com/questions/17924705/structure-of-arrays-vs-array-of-structures).
+The key concept of this paradigm is to structure data so that it maps
+closely to the form it will be represented and used within the relevant
+processing units. For example, it is a common mistake in object-oriented
+design to represent data as arrays of structures (AoS),
```
alanw0, line 166: It's not always a mistake. The preference for SoA vs AoS depends on the most computationally expensive algorithm, and which order the data needs to be accessed.

docs/dev_plan.md
```diff
@@ -0,0 +1,339 @@
...
+
+# Initialize test data
+all_points = [Point(random(i), random(i), random(i)) for i in range(100)]
+
+# Array of Structures
+AOS = [all_points[i].y for i in range(100)]
+
+# Structure of Arrays
+SOA = all_points.y + 10
+```
```
alanw0, line 201: Are we going to develop in python? Perhaps this example should be illustrated in whichever language we will use. Python array syntax is not very clear to anyone who doesn't already know python.

docs/dev_plan.md
```diff
@@ -0,0 +1,339 @@
...
+Access to the software is an important consideration for the
+longevity and relevance of OpenTurbine. In short, if the software is
+not accessible, it won't be used, extended, or maintained. All
+development efforts should always consider user and developer accessibility
+as a key driver. The distinction between "user" and "developer"
+is not always clear, but accessibility efforts should address concerns
+for both types of engagement with OpenTurbine.
+
+Documentation is the primary tool for addressing accessibility. As a
+guideline, the practice of documentation driven development should be
```
alanw0, line 218: Somewhere else in the document it said we would use test driven development. Documentation driven development is not the same and is vastly inferior.

docs/dev_plan.md
```diff
@@ -0,0 +1,339 @@
...
+widely adopted here. For any new feature or change, describe it first
+in the documentation. New features should be outlined in text to describe
+the high level objectives. Algorithmic development and design decisions
+should be included directly in the developer documentation. Before any
+code is written, the workflows in which the new code will be used should
+be expressed resulting in a proposed API. Finally, unit tests should be
+written to match the feature description in the documentation. Upon 
+implementation and subsequent iterations, the unit tests should retain
+their close correlation to the documentation so that any deviation
+between the implementation and documentation is caught through the tests.
```
alanw0, line 228: This is exactly the problem, unit-tests don't catch any deviation from documentation. How do you unit-test documentation? Unit-tests necessarily track the code. When any deviation happens, the code is the truth and the documentation is a lie.

docs/dev_plan.md
```diff
@@ -0,0 +1,339 @@
...
+- IEA 15-megawatt turbine where benchmark is a highly resolved shell-element Ansys model
+
+## Target baseline turbines
+
+- [NREL 5-megawatt reference turbine](https://www.nrel.gov/docs/fy09osti/38060.pdf)
+- [IEA 15-megawatt reference turbine](https://github.com/IEAWindTask37/IEA-15-240-RWT)
+
+## High-level development timeline
+
+**CY23 Q2**: The OpenTurbine team will implement a rigid-body dynamics solver following the
```
alanw0, line 316: Out of curiosity, why are these written in terms of CY (calendar year)? I've never seen planning or funding aligned with CY, always FY.

alanw0 (approved):
Overall seems pretty good, I just left a few nit-picky comments/questions.

docs/dev_plan.md
michaelasprague: @alanw0 -- the mapping right now is from the CFD surface to a line representation of the blade (or maybe even a point representation).  Nonetheless, we'll need a surface-to-surface coupling once shell/solid finite elements are used.

docs/dev_plan.md
rafmudaf: Good point @alanw0. In addition to being inaccurate, a description of the idea of data-oriented design is out of scope for this document. I've made some improvements to this section and removed much of the explanation based on your feedback.

docs/dev_plan.md
rafmudaf: It's a good point. I've removed this (see above comment), and we can add more meaningful examples in another section, if needed.

docs/dev_plan.md
rafmudaf: In this case, I intended that documentation driven development be adopted in addition to TDD rather than instead of it. I agree with your comment below that unit tests are the best way to document low-level code. For high-level code and public API's, though, I think we need to strike a balance of testing plus written documentation even at the risk of documentation becoming out of date. I see this as critical to ensuring that early design decisions and concepts are communicated to future developers so that they may understand when they're making fundamental changes. It also reduces the barrier to entry for future developers. Ultimately, the burden of ensuring that this documentation is accurate is on the development team who ideally treat the docs as a first-priority just like the code.

docs/dev_plan.md
rafmudaf: I see this was added in CY, then changed to FY, and again back to CY. @michaelasprague can you confirm the correct quarters for these milestones? If you think "FY" is unclear to anyone, It may be good to note that the FY is October through September.

docs/dev_plan.md
alanw0: Great points, thanks for the extra explanation. I totally agree that documentation is valuable, especially for high-level concepts and public APIs. It's definitely true that the developers need to prioritize keeping the docs up to date, because it doesn't happen automatically in the way that code and tests keep each other up to date.

Issue #22, opened 2023-05-16
Implement a prototype for the generalized-alpha solver


faisal-bhuiyan:
We would like to implement a prototype version of the generalized-alpha time integration scheme into OpenTurbine as part of the proof-of-concept (POC) study to model the dynamics of a rigid pendulum. To achieve this, we want to implement the algorithm presented in the [paper by Arnold and Bruls](https://link.springer.com/article/10.1007/s11044-007-9084-0). A basic description of the generalized-alpha method and the algorithm for constrained, nonlinear structural dynamics problems is provided in the [Overleaf document for the POC](https://www.overleaf.com/read/jrxxvrckpspb) (Section 5.1).

jjonkman:
@faisal-bhuiyan -- Overall, the generalized alpha scheme looks fine.  Can I suggest, though, that we you implement the Newton iterations in terms of incremental accelerations rather than incremental displacements?  I find this a more natural choice for structural dynamics problems and the Jacobians needed; this is also what we are using in the tight coupling algorithm being implemented now in OpenFAST.

In general, supporting algebraic constraints as proposed proposed looks fine.  But I would encourage the use of relative coordinates and algebraic elimination of constraints in the multi-body formulation of OpenTurbine as much as possible. The solution will be much more computationally efficient with fewer algebraic constraints.

michaelasprague:
@jjonkman Thanks for the comments and suggestion, and we will keep them under consideration.  At least for this proof-of-concept problem, we are interested in testing the displacement-based DAE-3 formulation, which is appealing for its simplicity.

PR #23, opened 2023-05-17
Solver architecture and generalized-alpha solver implementation

faisal-bhuiyan:
Addresses #22.

This PR needs a lot of development work before it will be ready for review, however, wanted to open a draft PR just to facilitate conversation around the solver design and get some early feedback.

Things to look for:
- Using [Eigen3](https://eigen.tuxfamily.org/index.php?title=Main_Page) as an external dependency to get a linear algebra library. This is what I am considering for the proof-of-concept study and will likely be replaced later on.
- Initial design of solver architecture.

TODO:
- Add unit tests
- Add private methods in `GeneralizedAlpha` class to create the `ResidualVector` and the `IterationMatrix`.
-


alanw0 (approved):
Faisal, this looks like a good start.
But I would suggest starting a unit-test first.
Even if the unit-test won't even compile yet and needs to be disabled or commented out, that's fine. But it will get you thinking about how the solver should be used.
Start with a test that feeds the solver an identity matrix and a trivial rhs, and the solver can initially just hard-code the solution to your trivial inputs.
Then as you add more unit-tests, the solver will progress to doing an actual solve.

alanw0:
In terms of your solver code, perhaps it is too early to be thinking about base-classes and inheritance design. Just start with the minimum needed to pass your first unit-test.
If you have thoughts about abstract solver classes and high-level design, express those in tests, and then that will drive you to create the needed code. The key to test-driven-development is picking/designing the right tests.

src/rigid_pendulum_poc/solver.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+GeneralizedAlphaSolver::GeneralizedAlphaSolver(
+    std::unique_ptr<LinearSystemSolver> linear_system_solver)
+    : linear_system_solver_(std::move(linear_system_solver)) {
+}
+
+void GeneralizedAlphaSolver::AlphaStep(const Eigen::VectorXd& gen_coords,
+                                       const Eigen::VectorXd& gen_coords_dot,
+                                       const Eigen::VectorXd& gen_coords_ddot,
+                                       const Eigen::VectorXd& accelaration) {
```
michaelasprague, line 15: "accelaration" is mispelled

src/rigid_pendulum_poc/solver.cpp
```diff
@@ -0,0 +1,62 @@
...
+    auto log = util::Log::Get();
+    log->Debug("AlphaStep\n");
+
+    // Some placeholders
+    float step_size = 0.1;
```
michaelasprague, line 20: time-step size should be an input

src/rigid_pendulum_poc/solver.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+    for (int i = 0; i < max_iterations; i++) {
+        // Compute the residuals
+        // TODO Define a function to compute the residuals
+        Eigen::VectorXd r1 = gen_coords_next - gen_coords - step_size * gen_coords_dot -
+                             step_size * step_size * (0.5 - beta) * accelaration;
+        Eigen::VectorXd r2 =
+            gen_coords_dot_next - gen_coords_dot - step_size * (1 - gamma) * accelaration;
+
+        // Compute the jacobian
```
michaelasprague, line 45: Not sure what this is; there isn't a Jacobian, per say.  There is just the "iteration matrix" which will be called below.

src/rigid_pendulum_poc/solver.cpp
```diff
@@ -0,0 +1,62 @@
...
+        Eigen::MatrixXd jacobian {};
+        // TODO Define a function to compute the jacobian
+        // jacobian = jacobian - step_size* step_size* (0.5 - beta) * alpha_f;
+
+        // Compute the iteration matrix
+        // TODO Define a function to compute the iteration matrix
+        Eigen::MatrixXd iteration_matrix {};
+        // iteration_matrix = iteration_matrix - step_size* step_size* (0.5 - beta) * alpha_m;
```
michaelasprague, line 53: Not sure what this is.  The "Iteration Matrix" is S_t in Arnold and Bruls (2007), and it is the matrix that will be passed to the linear system solver at each Newton iteration.

src/rigid_pendulum_poc/solver.cpp
```diff
@@ -0,0 +1,62 @@
...
+    Eigen::VectorXd accelaration_next =
+        (1 / (1 - alpha_m)) * (alpha_f * gen_coords_ddot - alpha_m * accelaration);
+    gen_coords_next = gen_coords_next + step_size * step_size * beta * accelaration_next;
+    gen_coords_dot_next = gen_coords_dot_next + step_size * beta * accelaration_next;
+    Eigen::VectorXd gen_coords_ddot_next {};
+
+    for (int i = 0; i < max_iterations; i++) {
+        // Compute the residuals
+        // TODO Define a function to compute the residuals
+        Eigen::VectorXd r1 = gen_coords_next - gen_coords - step_size * gen_coords_dot -
```
michaelasprague, line 40: Are r1 and r2 just placeholders?  I would assume there would be a call to a function to create these residual vectors.  Also, there should be a check after they are formed to see if the tolerance has been reached.

michaelasprague (commented):
A few comments about the algorithm structure.

PR #24, opened 2023-05-19
Implement a solver for linear systems based on Kokkos data structures and LAPACK

faisal-bhuiyan:
This is WIP. Wanted to make sure we are on the right track using test-driven development as the development principle and relying on `Kokkos` and `Kokkos-Kernels` for container and linear algebra support.


faisal-bhuiyan:
I was able to add `kokkos-kernels` as a dependency for OpenTurbine and started off with the TDD we talked about in the last meeting. However, I am not sure about how to get a linear solver working based on `Kokkos-Kernels` and would appreciate your thoughts. Particularly, I don't see any explicit documentation regarding wrappers for LAPACK methods and I don't want to write linear algebra library for our work. There must be something already implemented in Kokkos ecosystem, right?

tests/unit_tests/test_solver.cpp
```diff
@@ -0,0 +1,59 @@
...
+    }
+};
+
+TEST_F(LinearSolverTest, Test1) {
+    LinearSolver solver {};
+    EXPECT_EQ(solver.GetSolverType(), SolverType::kNone);
+
+    Kokkos::View<double*> x("x", 1);
+    x[0] = 1.0;
+    Kokkos::View<double*> A("A", 1);
```
ddement, line 50: A is a matrix, so it should be a `Kokkos::View<double**>`

tests/unit_tests/test_solver.cpp
```diff
@@ -0,0 +1,59 @@
...
+    A[0] = 1.0;
+    Kokkos::View<double*> b("b", 1);
+    Kokkos::deep_copy(b, solver.Solve(A, x));
```
ddement, line 53: We shouldn't need a deep copy here, because Kokkos views behave like smart pointers.  We may want to pass in the output vector so that we can have better control on memory allocation (i.e. so `solver.Solve` doesn't have to do this allocation)

src/rigid_pendulum_poc/solver.h
```diff
@@ -0,0 +1,33 @@
...
+};
+
+/// @brief A class for solving linear systems
+class LinearSolver {
+public:
+    LinearSolver(SolverType=SolverType::kNone);
+
+    SolverType GetSolverType() const { return solver_type_; }
+
+    Kokkos::View<double*> Solve(const Kokkos::View<double*>&, const Kokkos::View<double*>&) const;
```
ddement, line 25: There's no need to pass `Kokkos::View` by reference - their smart-pointer like semantics mean that copies are very cheap and are actually safer with regards to object lifetime.

src/rigid_pendulum_poc/solver.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include <iostream>
+#include <vector>
+
+#include <Kokkos_Core.hpp>
+
+namespace openturbine::rigid_pendulum {
+
+/// @brief Enumerate different types of solvers
+enum class SolverType {
```
ddement, line 11: Do we need to define different solver types yet?  From a TDD perspective, we don't have a use for this functionality yet, so we should hold off on implementing it until we have a use case for it (as demonstrated by tests).

src/rigid_pendulum_poc/solver.cpp
```diff
@@ -0,0 +1,23 @@
+#include "src/rigid_pendulum_poc/solver.h"
+
+#include <Kokkos_Core.hpp>
+#include <KokkosBlas1_abs.hpp> // For absolute value operations
+#include <KokkosBlas1_axpby.hpp> // For linear combination operations
+#include <KokkosBlas1_dot.hpp> // For dot product operations
+#include <KokkosBlas1_nrm2.hpp> // For norm operations
```
ddement, line 7: No need to include these until we need them.

src/rigid_pendulum_poc/solver.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
```
ddement, line 1: While I like `#pragma once`, we need to make sure that it is supported on all platforms that we'll be targeting

tests/unit_tests/test_solver.cpp
```diff
@@ -0,0 +1,59 @@
+#include "gtest/gtest.h"
+
+#include <Kokkos_Core.hpp>
+
+#include "src/rigid_pendulum_poc/solver.h"
+
+namespace oturb_tests {
```
ddement, line 7: perhaps this should be `openturbine_tests`?  We won't be calling this namespace much, if at all, so it can be longer without sacrificing line length.

tests/unit_tests/test_solver.cpp
```diff
@@ -0,0 +1,59 @@
...
+    EXPECT_LT(SolverType::kNone, SolverType::kIterativeLinearSolver);
+    EXPECT_EQ(SolverType::kIterativeLinearSolver, SolverType::kIterativeLinearSolver);
+    EXPECT_NE(SolverType::kIterativeLinearSolver, SolverType::kDirectLinearSolver);
+    EXPECT_NE(SolverType::kIterativeLinearSolver, SolverType::kNone);
+    EXPECT_LT(SolverType::kNone, SolverType::kIterativeNonlinearSolver);
+    EXPECT_EQ(SolverType::kIterativeNonlinearSolver, SolverType::kIterativeNonlinearSolver);
+    EXPECT_NE(SolverType::kIterativeNonlinearSolver, SolverType::kIterativeLinearSolver);
+    EXPECT_NE(SolverType::kIterativeNonlinearSolver, SolverType::kDirectLinearSolver);
+    EXPECT_NE(SolverType::kIterativeNonlinearSolver, SolverType::kNone);
+}
```
ddement, line 31: This test puts a contraint on our enum ordering.  Isn't the usual use case just to check for equality (which is a built-in functionality, similar to checking integer equality)?  If we don't need to know the ordering, or underlying number representing the enum, I don't know if we should add this constraint yet.

ddement (commented):
We may want to start testing from a more top-down perspective.  I'm thinking a test for a base-case of the generalize alpha solver.  Then, as we quickly need more functionality, we can add things like the linear solver, coordinate manipulations, etc.  This might help us to add only the features we need for now to solve our simple problem.

faisal-bhuiyan:
> `kokkos-kernels` provides a `KokkosBlas::gesv` method that performs a simple dense LU solve, if that works for this application. `kokkos-kernels` also has a lot of dense solvers under the `KokkosBatched` namespace. These are mainly used for when you need to solve a lot of small systems, so you can saturate the GPU better, but you should be able to use them for a single system just as well.
> 
> I do admit to being surprised that they don't have an easier LAPACK type interface in there. I understand that those algorithms are harder to parallelize, but it still seems like it would be a good tool in the toolbox.

Doesn't look there they implemented the `KokkosBlas::gesv` as of yet. Am I missing something here?

https://github.com/kokkos/kokkos-kernels/blob/develop/blas/impl/KokkosBlas_gesv_impl.hpp

src/rigid_pendulum_poc/solver.h
alanw0: We've considered using pragma once in our stk code-base. So far we haven't, only because it isn't technically standard, last time we checked. It does seem to be supported by every platform we know of though, so it's probably safe. But we're pretty cautious so we haven't adopted it yet. I could go either way in this case.

alanw0:
> @ddement Doesn't look like they implemented the `KokkosBlas::gesv` method as of yet. Am I missing something here?

It looks like you're correct, that's puzzling. I wonder why they would leave an un-implemented stub in there like that...

ddement:
> > `kokkos-kernels` provides a `KokkosBlas::gesv` method that performs a simple dense LU solve, if that works for this application. `kokkos-kernels` also has a lot of dense solvers under the `KokkosBatched` namespace. These are mainly used for when you need to solve a lot of small systems, so you can saturate the GPU better, but you should be able to use them for a single system just as well.
> > I do admit to being surprised that they don't have an easier LAPACK type interface in there. I understand that those algorithms are harder to parallelize, but it still seems like it would be a good tool in the toolbox.
> 
> @ddement Doesn't look like they implemented the `KokkosBlas::gesv` method as of yet. Am I missing something here?
> 
> https://github.com/kokkos/kokkos-kernels/blob/develop/blas/impl/KokkosBlas_gesv_impl.hpp

That's really odd.  They do seem to have one based on TPLs:
https://github.com/kokkos/kokkos-kernels/blob/develop/blas/tpls/KokkosBlas_gesv_tpl_spec_decl.hpp

Basically, it is a wrapper for the system BLAS gesv, so we don't have to do it ourselves.  They also let you use a device based solver TPL.  We just have to make sure that we build Kokkos kernels with the right BLAS.

Another way forward would be to write our own wrapper around a BLAS/LAPACK TPL that takes in the Kokkos views.  My first instinct says this would be harder on us, but I'll leave that judgment up to you.

ddement:
> > We may want to start testing from a more top-down perspective. I'm thinking a test for a base-case of the generalize alpha solver. Then, as we quickly need more functionality, we can add things like the linear solver, coordinate manipulations, etc. This might help us to add only the features we need for now to solve our simple problem.
> 
> @ddement While I completely agree with the top-down approach and see the value of using this viewpoint as we build the minimum viable proof-of-concept for the rigid body pendulum, I am in favor of implementing the required features for the generalized-alpha solver separately. This will keep the development work in focused, relatively small (few hundred lines of change max), reviewer-friendly pull requests (as opposed to a large PR containing several Ks of lines of code). I do see ourselves merging back to the [draft pull request on generalized-alpha solver](https://github.com/Exawind/openturbine/pull/23) and completing the GA solver incrementally.
> 
> Does the above plan sound like a reasonable approach to you?

That sounds like a plan to me

ddement:
@faisal-bhuiyan You'll have to configure kokkos-kernels to use BLAS.  If you're using CMAKE directly to build it, you can just add ` -DKokkosKernels_ENABLE_TPL_BLAS=ON`.  If you're using Spack, there should be a corresponding flag.  It should be able to find the packages that are loaded into your path - I don't know what your development environment looks like to know what modules you need for this.  

OpenBLAS should be as good of a BLAS as any for now.  I know that CMAKE has a findBLAS function, as well as using find_package, so that should give us some modularity when including/linking to BLAS ourselves.

faisal-bhuiyan:
> @faisal-bhuiyan You'll have to configure kokkos-kernels to use BLAS. If you're using CMAKE directly to build it, you can just add ` -DKokkosKernels_ENABLE_TPL_BLAS=ON`. If you're using Spack, there should be a corresponding flag. It should be able to find the packages that are loaded into your path - I don't know what your development environment looks like to know what modules you need for this. Let me know if you need help getting this configured.

I used spack to install kokkos-kernels as follows: `spack install kokkos-kernels +blas`. Does it seem reasonable? That's what is provided in their build instructions [here](https://github.com/kokkos/kokkos-kernels/blob/master/BUILD.md#spack).

> OpenBLAS should be as good of a BLAS as any for now. I know that CMAKE has a findBLAS function, as well as using find_package, so that should give us some modularity when including/linking to BLAS ourselves.

I agree and that's what is installed as a dependency for kokkos-kernels with TPL BLAS as pointed out above. I used a `find_package(BLAS REQUIRED)` into CMake and got the following output:

```
-- OpenTurbine Information:
-- VERSION = v0.0.0-69-gc27d7cd-DIRTY
-- CMAKE_SYSTEM_NAME = Darwin
-- CMAKE_CXX_COMPILER_ID = AppleClang
-- CMAKE_CXX_COMPILER_VERSION = 14.0.3.14030022
-- CMAKE_BUILD_TYPE = RELEASE
-- Enabled Kokkos devices: SERIAL
-- Looking for sgemm_
-- Looking for sgemm_ - not found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Looking for dgemm_
-- Looking for dgemm_ - found
-- Found BLAS: /Library/Developer/CommandLineTools/SDKs/MacOSX13.3.sdk/System/Library/Frameworks/Accelerate.framework  
-- Enabled Kokkos devices: SERIAL
-- Found GTest: /opt/homebrew/lib/cmake/GTest/GTestConfig.cmake (found version "1.13.0")  
-- Enabled Kokkos devices: SERIAL
-- Enabled Kokkos devices: SERIAL
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/fbhuiyan/dev/openturbine/build
```

I still get the same error when I run the main function with call to `KokkosBlas::gesv`. Any thoughts on next steps here?

src/rigid_pendulum_poc/solver.h
faisal-bhuiyan: Thanks for the helpful pointers on Kokkos::View here and the following comments. Clearly, I am just getting started with this container and will need to spend a little more time with it to get to the point of idiomatic use.

tests/unit_tests/test_solver.cpp
faisal-bhuiyan: Sounds like a good idea to me!

faisal-bhuiyan:
@ddement @alanw0 @michaelasprague  I was able to ask some questions on kokkos slack and was referred to the GMRES solver (an iterative solver for sparse matrices) by a kokkos-kernels developer. Please take a look in the following link.

https://github.com/kokkos/kokkos-kernels/tree/master/example/gmres

Before I dive into any implementation for OpenTurbine, want to make sure what you might think of the suitability of this particular solver for our purposes.

faisal-bhuiyan:
> This code is not production ready. It lacks guard rails to prevent using device memory or non-row-major matrices. It also lacks size checking and error handling for in input system. These should be added before merging into master to prevent segfault risk.

@ddement I added some of the checks and error handling to the solver function. Check for non-row major matrices and more than one r.h.s does not appear to be necessary at this point since the function signature does not allow them and throws. I am not sure about the device memory stuff - could explain that a little more?

Also, could you take a quick scan and let me know if there is anything else we might need to add to this PR before going into review?

src/rigid_pendulum_poc/solver.h
```diff
@@ -1,10 +1,17 @@
...
-[[nodiscard]] int solve_linear_system(Kokkos::View<double**> system, Kokkos::View<double*> solution);
+/// @brief Solve a linear system of equations using LAPACKE's dgesv
+/// @details This function solves a linear system of equations using LAPACKE's
+///     dgesv function. The system is of the form Ax = b, where A is a square matrix
+///     (nxn) of coefficients, x is a vector (nx1) of unknowns, and b is a vector
+///     (nx1) of right-hand side values. The solution is stored in the vector b.
+/// @param system A matrix of coefficients
+/// @param solution A vector of right-hand side values
+/// @return An integer indicating the success of the solution
+int solve_linear_system(Kokkos::View<double**>, Kokkos::View<double*>);
```
ddement, line 15: Do we want to keep nodiscard?  It is a not so gentle reminder for the caller to check the error value.  Or we could check the lapack error value and throw inside the function if it is anything other than 0.

tests/unit_tests/test_solver.cpp
```diff
@@ -1,135 +1,172 @@
...
 }
 
-Kokkos::View<double*, Kokkos::DefaultHostExecutionSpace>
-create_vector(std::vector<double> values)
-{
-  auto vector = Kokkos::View<double*, Kokkos::DefaultHostExecutionSpace>("vector", values.size());
-  auto entries = Kokkos::RangePolicy<Kokkos::DefaultHostExecutionSpace>(0, values.size());
-  auto fill_vector = [vector, values](int index) { vector(index) = values[index]; };
+Kokkos::View<double*, Kokkos::DefaultHostExecutionSpace> create_vector(
+    const std::vector<double>& values) {
```
ddement, line 27: This formatting is not great.  We should change our clang-format to make it better (I think putting the line break after the type instead of in the argument list is more readable).  Most of the format changes are just stylistic differences, but ones like this risk making our production code harder to read, in my opinion

tests/unit_tests/test_solver.cpp
```diff
@@ -1,135 +1,172 @@
...
-  EXPECT_NEAR(solution(2), exactSolution(2), 10*std::numeric_limits<double>::epsilon());
+TEST(LinearSolverTest, CheckMatrixShape) {
+    // Try to solve a non-square system
+    auto matrix_2x3 = create_matrix({{1., 2., 3.}, {4., 5., 6.}});
+    auto solution_2x1 = create_vector({1., 1.});
+
+    EXPECT_THROW(openturbine::rigid_pendulum::solve_linear_system(matrix_2x3, solution_2x1),
+                 std::invalid_argument);
+
+    // Try to solve another non-square system
```
ddement, line 143: I'd make this two separate tests

faisal-bhuiyan: I am not well-versed in the standard practices of TDD, so pardon my ignorance here. Do we always need to write separate tests every time we introduce new inputs? To be specific, for the tests here they are checking if the matrix is square or not - is it not recommended to assert several matrices even though you reach the solution in production code in the first pass?

ddement:
Basically, we don't want to try to use a view of device memory here, because LAPACK is host only.  Thinking about it, the easiest way would probably be to put the `Kokkos::DefaultHostExecutionSpace` in the type signature.  We could also do some templates and type checking magic, but that's probably too complicated for this stage in the game.

src/rigid_pendulum_poc/solver.h
ddement: Sounds good to me

tests/unit_tests/test_solver.cpp
ddement: In general, I like to make every set of inputs its own test. This keeps our tests really simple and easy to understand. Also, a failure points you right to the failure points and there are very few moving parts to complicate things.  

You should definitely add as many tests as reasonable, even once they all pass, but keeping them separate aids in writing them now and future reading and debugging.  Similar to how you want your functions to have a single responsibility.

faisal-bhuiyan:
@ddement I meant to ask you, did you need to do anything in CMake related to finding the LAPACKE package such as `find_package(LAPACKE REQUIRED)` or something similar? I had to do something non-trivial for CMake to find LAPACKE and wanted to make sure if there's a better way to do it.

tests/unit_tests/test_solver.cpp
ddement: Fair enough. I'll play around with it and figure something out in parallel. We can probably move forward with this PR.

ddement:
> @ddement I meant to ask you, did you need to do anything in CMake related to finding the LAPACKE package (pertaining to the `#include <lapacke.h>` in solver) such as `find_package(LAPACKE REQUIRED)` or something similar? I had to do something non-trivial for CMake to be able to find LAPACKE and wanted to make sure if there's a better way to do it.

I only needed to do the find_package command you listed in the src/CMakeLists.txt.  We should only at most need to point CMake to where it lives on a given machine at cmake time, unless it already lives in your path as it does on my machine.  That being said, I may be missing something.  I've never been confused for a CMake expert.

faisal-bhuiyan:
@michaelasprague @alanw0 @ddement This one is ready for a final review at this point. I will need to make some updates to the [documentation regarding compiling OpenTurbine](https://exawind.github.io/openturbine/compiling.html) since we added a dependency on LAPACK. Will work with @rafmudaf to figure that out since I don't have the write access to it.

tests/unit_tests/test_solver.cpp
rafmudaf, line 1: It may be good to rename this to something like `test_pendulum_solver` since there may end up being multiple solver-files in the short term.

rafmudaf (approved):
I reviewed the code infrastructure (CMake, CI, docs), and that part looks good to me.

alanw0 (approved):
What's the difference between lapack and lapacke?
Is lapacke portable, widely available?

PR #25, opened 2023-05-24
Implement linear solver using Kokkos data structures and LAPACK

ddement:
This set of commits provides a wrapper for LAPACK's general solver using `Kokkos::View`s.  
This would usually be just one commit, but many intermediate commits have be provided to demonstrate the TDD process used to arrive at this solution.
This code is not production ready.  It lacks guard rails to prevent using device memory or non-row-major matrices.  It also lacks size checking and error handling for in input system.  These should be added before merging into master to prevent segfault risk.


faisal-bhuiyan:
Thanks for adding the wrapper for LAPACK's general solver, @ddement. I will add the additional stuff necessary for this feature to be merged into main.

PR #30, opened 2023-06-07
Implement the framework of a genearlized-alpha solver based on TDD

faisal-bhuiyan:
Implements #22.

Framework of the fully nonlinear generalized-alpha solver was implemented in this work following the principles of test-driven development. To this end, a `GeneralizedAlphaTimeIntegrator` class was introduced to handle the time stepping, integration, and Newton-Raphson iterations. 

The implementation is problem agnostic, meaning it is general enough to be used on any structural dynamics problem. We need to implement additional features for this to handle the problem of rigid pendulum consisting of quaternions as generalized coordinates.

### DIFF
--- a/src/rigid_pendulum_poc/solver.cpp
+++ b/src/rigid_pendulum_poc/solver.cpp
@@ -109,9 +109,12 @@ State operator+=(State& lhs, const State& rhs) {
 }
 
 GeneralizedAlphaTimeIntegrator::GeneralizedAlphaTimeIntegrator(
-    double initial_time, double time_step, size_t number_of_steps
+    double initial_time, double time_step, size_t number_of_steps, bool nonlinear_analysis
 )
-    : initial_time_(initial_time), time_step_(time_step), number_of_steps_(number_of_steps) {
+    : initial_time_(initial_time),
+      time_step_(time_step),
+      number_of_steps_(number_of_steps),
+      nonlinear_analysis_(nonlinear_analysis) {
     this->current_time_ = initial_time;
 }
 
@@ -140,7 +143,10 @@ State GeneralizedAlphaTimeIntegrator::AlphaStep(const State& state) {
     auto [linear_coords, linear_velocity, algo_acceleration] =
         UpdateLinearSolution(gen_coords, gen_velocity, gen_accln, algo_accln);
 
-    // TODO: Implement nonlinear update
+    // Perform Newton-Raphson iterations to perform nonlinear part of generalized alpha method
+    if (this->nonlinear_analysis_) {
+        // TODO: Implement nonlinear update
+    }
 
     return state;
 }
--- a/src/rigid_pendulum_poc/solver.cpp
+++ b/src/rigid_pendulum_poc/solver.cpp
@@ -156,6 +156,16 @@ State GeneralizedAlphaTimeIntegrator::UpdateLinearSolution(const State& state) {
         }
     );
 
+    auto log = util::Log::Get();
+    log->Debug("Linear solution: gen_coords, gen_velocity, algo_acceleration\n");
+
+    for (size_t i = 0; i < size; i++) {
+        log->Debug(
+            "row " + std::to_string(i) + ": " + std::to_string(gen_coords_next(i)) + "\t" +
+            std::to_string(gen_velocity_next(i)) + "\t" + std::to_string(algo_accln_next(i)) + "\n"
+        );
+    }
+
     return State(gen_coords_next, gen_velocity_next, gen_accln, algo_accln_next);
 }
 
--- a/src/rigid_pendulum_poc/solver.cpp
+++ b/src/rigid_pendulum_poc/solver.cpp
@@ -52,22 +52,22 @@ void solve_linear_system(HostView2D system, HostView1D solution) {
 }
 
 State::State()
-    : generalized_coords_("generalized_coords", 1),
-      generalized_coords_dot_("generalized_coords_dot", 1),
-      generalized_coords_dot_dot_("generalized_coords_dot_dot", 1),
-      accelerations_("accelerations", 1) {
+    : generalized_coords_("generalized_coordinates", 1),
+      generalized_velocity_("generalized_velocity", 1),
+      generalized_accelerations_("generalized_accelerations", 1),
+      algorithmic_accelerations_("algorithmic_accelerations", 1) {
 }
 
-State::State(HostView1D gen_coords, HostView1D gen_coords_dot, HostView1D gen_coords_ddot,
-             HostView1D accls)
-    : generalized_coords_("generalized_coords", gen_coords.size()),
-      generalized_coords_dot_("generalized_coords_dot", gen_coords_dot.size()),
-      generalized_coords_dot_dot_("generalized_coords_dot_dot", gen_coords_ddot.size()),
-      accelerations_("accelerations", accls.size()) {
+State::State(HostView1D gen_coords, HostView1D gen_velocity, HostView1D gen_accln,
+             HostView1D algo_accln)
+    : generalized_coords_("generalized_coordinates", gen_coords.size()),
+      generalized_velocity_("generalized_velocity", gen_velocity.size()),
+      generalized_accelerations_("generalized_accelerations", gen_accln.size()),
+      algorithmic_accelerations_("algorithmic_accelerations", algo_accln.size()) {
     Kokkos::deep_copy(generalized_coords_, gen_coords);
-    Kokkos::deep_copy(generalized_coords_dot_, gen_coords_dot);
-    Kokkos::deep_copy(generalized_coords_dot_dot_, gen_coords_ddot);
-    Kokkos::deep_copy(accelerations_, accls);
+    Kokkos::deep_copy(generalized_velocity_, gen_velocity);
+    Kokkos::deep_copy(generalized_accelerations_, gen_accln);
+    Kokkos::deep_copy(algorithmic_accelerations_, algo_accln);
 }
 
 HostView1D operator+(const HostView1D& view1, const HostView1D& view2) {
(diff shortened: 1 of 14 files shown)


src/rigid_pendulum_poc/solver.h
```diff
@@ -4,14 +4,141 @@
 
 namespace openturbine::rigid_pendulum {
 
+using HostView1D = Kokkos::View<double*, Kokkos::DefaultHostExecutionSpace>;
+using HostView2D = Kokkos::View<double**, Kokkos::DefaultHostExecutionSpace>;
```
ddement, line 8:
```suggestion
using HostView1D = Kokkos::View<double*, Kokkos::HostSpace>;
using HostView2D = Kokkos::View<double**, Kokkos::HostSpace>;
```

ddement (commented):
This should cover everything we discussed yesterday plus other small things.  Looks like we're heading in the right direction.

src/rigid_pendulum_poc/solver.cpp
```diff
@@ -52,4 +51,112 @@ void solve_linear_system(Kokkos::View<double**, Kokkos::DefaultHostExecutionSpac
...
+    }
+
+    log->Debug("Time integration completed successfully \n");
+}
+
+void GeneralizedAlphaTimeIntegrator::AlphaStep() {
+    // Perform the linear update
+    auto gen_coords = this->state_.GetGeneralizedCoordinates();
+    auto gen_coords_dot = this->state_.GetGeneralizedCoordinatesDot();
+    auto gen_coords_ddot = this->state_.GetGeneralizedCoordinatesDotDot();
```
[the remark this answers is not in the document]
faisal-bhuiyan: Nope, velocity and accelerations are much better names. Thanks for pointing it out.

faisal-bhuiyan:
Update: At this point, we have the nonlinear solver framework implemented with dummy implementations for iteration matrix and the residual function. I will shift focus on deriving the terms for the rigid pendulum problem.

alanw0 (approved):
I think this is looking pretty good.

src/rigid_pendulum_poc/solver.cpp
```diff
@@ -52,4 +54,242 @@ void solve_linear_system(Kokkos::View<double**, Kokkos::DefaultHostExecutionSpac
...
+        auto [nonlinear_coords, nonlinear_velocity, nonlinear_acceleration] =
+            UpdateNonLinearSolution(gen_coords, gen_velocity, gen_accln);
+    }
+
+    return state;
+}
+
+std::tuple<HostView1D, HostView1D, HostView1D> GeneralizedAlphaTimeIntegrator::UpdateLinearSolution(
+    const HostView1D& gen_coords, const HostView1D& gen_velocity, const HostView1D& gen_accln,
+    const HostView1D& algo_accln
```
ddement, line 158: Pass views by value, not by reference

src/rigid_pendulum_poc/solver.cpp
```diff
@@ -52,4 +54,242 @@ void solve_linear_system(Kokkos::View<double**, Kokkos::DefaultHostExecutionSpac
...
+            algo_accln(i) = lhs_algo_accln(i) + rhs_algo_accln(i);
+        }
+    );
+
+    return State(gen_coords, gen_velocity, gen_accln, algo_accln);
+}
+
+State operator+=(State& lhs, const State& rhs) {
+    return lhs = lhs + rhs;
+}
```
ddement, line 109: Do we still need these operators?

ddement (approved):

src/rigid_pendulum_poc/solver.cpp
```diff
@@ -57,4 +54,190 @@ void solve_linear_system(
...
+    }
+
+    return {gen_coords, gen_velocity, algo_accln};
+}
+
+HostView1D GeneralizedAlphaTimeIntegrator::ComputeResiduals(HostView1D forces) {
+    // TODO: Compute the residuals
+    // r^q = M(q) * q'' - f + phi_q^T * lambda
+    // This is a just a placeholder, returns the provided forces as the residuals
+    return forces;
```
ddement, line 200: If we return a vector of all ones here (of the same size as forces), it should make testing really simple.  

Once this testing is in place, you can abstract this functionality out to an object that has its own "getResidual" and "getInterationMatrix" functions.

src/rigid_pendulum_poc/solver.cpp
```diff
@@ -57,4 +54,190 @@ void solve_linear_system(
...
+    Kokkos::parallel_for(
+        size,
+        KOKKOS_LAMBDA(const int i) {
+            // Update algorithmic acceleration once soln has converged
+            algo_accln(i) += (1 - kALPHA_F) / (1 - kALPHA_M) * gen_accln(i);
+        }
+    );
+
+    return State(gen_coords, gen_velocity, gen_accln, algo_accln);
+}
```
ddement, line 159: This structure of this function looks good (assuming that the math is correct).  Let's focus on ComputeResudials and ComputeIterationMatrix as our customization points.  They should eventually call out to a function object that has specific information for how to do this for our problem or simple objects used for testing.  

We may want to be able to simplify the alpha, beta, and gamma values for initial testing (set them to 1 or 0, depending on what makes the math work out simplest).  This should help to show that they are implemented correctly and simplify our initial testing of the AlphaStep function

src/rigid_pendulum_poc/solver.cpp
faisal-bhuiyan: > If we return a vector of all ones here (of the same size as forces), it should make testing really simple.

This sounds like a reasonable approach to me, with following caveat.

We are always going to have either convergence/non-convergence ([the check at line 129 here](https://github.com/Exawind/openturbine/blob/715156ff17465c8dd966db176275cbda81d83da6/src/rigid_pendulum_poc/solver.cpp#L129)) for the `AlphaStep` based on the hard-coded behavior here. That would prevent us from observing the convergence of the `AlphaStep` after, let's say, one iteration. Any thoughts on how to test out the behavior of the GA solver without implementing the details of the `ComputeResiduals()` function?

FWIW, my opinion is to return non-convergence based on the hard-coded residual and iteration matrix and test the expected solution after one iteration of the GA time integration by setting the [max_iterations](https://github.com/Exawind/openturbine/blob/715156ff17465c8dd966db176275cbda81d83da6/src/rigid_pendulum_poc/solver.h#L52) at 1. What do you think?

src/rigid_pendulum_poc/solver.cpp
faisal-bhuiyan: > This structure of this function looks good (assuming that the math is correct). Let's focus on ComputeResudials and ComputeIterationMatrix as our customization points. They should eventually call out to a function object that has specific information for how to do this for our problem or simple objects used for testing.

As discussed above, this sounds like a very reasonable approach to me.

> We may want to be able to simplify the alpha, beta, and gamma values for initial testing (set them to 1 or 0, depending on what makes the math work out simplest). This should help to show that they are implemented correctly and simplify our initial testing of the AlphaStep function

Yes, this is also a reasonable approach. We do need to provide some guard rails around the values of the constants, I will implement that as we get done with the testing part of `AlphaStep()`.

src/rigid_pendulum_poc/solver.cpp
ddement: I think the first step is to use a simple system that will converge after one nonlinear iteration.  This is easy to both formulate and test.  I'd first start with an identity iteration matrix like you have.  Once that's working, you can maybe put together a non-singular, non-identity iteration matrix that you know the exact solution for to ensure everything is working for a matrix that isn't just the identity.  The same logic follows for the residual - start with something simple like all ones, then we can do a static residual vector that is a little more complex.

The nonlinear case is a little harder to implement the residual/iteration matrix.  Once you have those, I agree with setting max_iterations to 1 to test the behavior (since you can easily compute the result by hand).  I'd even set max_iterations to 2 to test the values after that to compare to a hand computed value (assuming that the system is chosen so that it is possible).  Once you know it works to 2 iterations, it should work for all greater numbers.  

The convergence check itself might be harder to test, but we can cross that bridge when we get to it.

tests/unit_tests/rigid_pendulum_poc/test_generalized_alpha_solver.cpp
```diff
@@ -0,0 +1,246 @@
...
+    EXPECT_EQ(time_integrator.GetMaxIterations(), 17);
+}
+
+TEST(TimeIntegratorTest, AlphaStepSolutionAfterOneIncWithZeroAcceleration) {
+    auto initial_state = State();
+
+    expect_kokkos_view_1D_equal(initial_state.GetGeneralizedCoordinates(), {0.});
+    expect_kokkos_view_1D_equal(initial_state.GetGeneralizedVelocity(), {0.});
+    expect_kokkos_view_1D_equal(initial_state.GetGeneralizedAcceleration(), {0.});
+    expect_kokkos_view_1D_equal(initial_state.GetAlgorithmicAcceleration(), {0.});
```
ddement, line 202: This is tested elsewhere, so there's no need to put these checks here.  Removing unnecessary checks makes your tests shorter and easier to read, plus it ensures that you can easily see what new features are exercised by this test.

src/rigid_pendulum_poc/solver.h
```diff
@@ -4,14 +4,138 @@
...
+    /// Returns the gamma parameter
+    inline double GetGamma() const { return kGAMMA_; }
+
+    /// Returns the maximum number of iterations
+    inline size_t GetMaxIterations() const { return kMAX_ITERATIONS_; }
+
+    /// Performs the time integration and returns a vector of States over the time steps
+    std::vector<State> Integrate(
+        const State&, std::function<HostView2D(size_t)> iteration_matrix = create_identity_matrix,
+        std::function<HostView1D(size_t)> residual_vector = create_identity_vector
```
ddement, line 91: This should work to let us customize these calculations.  

Usually, these function objects would be passed in at construction time and stored, usually as a single class that defines two different methods.  However, let's continue with what you have now and see how it looks in the end.  

We might also need to revisit if the `std::function<...>` should be passed by reference.  Type erasure (the technique used to implement it) is complex, so reference vs value may have both performance and correctness implications.

ddement:
We should merge this pull request sooner rather than later.  It is getting hard to see what the new changes are with each commit set.  We can then continue development using a new pull request.

src/rigid_pendulum_poc/solver.h
faisal-bhuiyan: Sounds good to me.

faisal-bhuiyan:
> We should merge this pull request sooner rather than later. It is getting hard to see what the new changes are with each commit set. We can then continue development using a new pull request.

I agree. This can be considered as a completed feature and we can move on to merge this upon review. Proposed enhancements (such as mapping and quaternion-based residual/iteration matrices) can be performed in separate pull request(s).

ddement (approved):

alanw0:
I agree. It's good that you are getting some code and tests fleshed out.
Perhaps David and I can ramp up more work on this... Perhaps we can help more with abstractions and restructuring for inserting test inputs vs "real" inputs, etc., while you move on with more formulation/algorithm work?

alanw0:
I would suggest going ahead and merging this one, and we can work in new pull requests.

faisal-bhuiyan:
> I would suggest going ahead and merging this one, and we can work in new pull requests.

Could you review this one more time and approve if looks good? Apologies for having to make you do this multiple times.

alanw0 (approved):

PR #31, opened 2023-06-27
Minor update to the clang-format file for better readability

faisal-bhuiyan:
Modified .clang-format and applied it to `src/` and `tests/ `directory files. Highlights are as follows
- Added block indent for alignment after open bracket
- Increased the column limit to 101
- Fixed the include categories to improve grouping

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -8,6 +8,12 @@
 # the class/struct keyword
 AccessModifierOffset: -4
 
+# Bracket Alignment Style: Always break after an open bracket, if the
+# parameters don’t fit on a single line. Closing brackets will be placed
+# on a new line.
+# Applies to round (i.e. parentheses), square, and angle brackets
+AlignAfterOpenBracket: BlockIndent
+
 # Do NOT allow short enums on a single line
 AllowShortEnumsOnASingleLine: false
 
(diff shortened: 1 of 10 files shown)


.clang-format
```diff
@@ -28,8 +34,8 @@ AllowShortLoopsOnASingleLine: false
 # in this project
 BasedOnStyle: Google
 
-# Increase the column limit to 100
-ColumnLimit: 100
+# Increase the column limit to 101
+ColumnLimit: 101
```
ddement, line 38: It looks like we can set this to 0 and it won't change the line length from what you type.  This should make developing good looking code faster than having to tweak this term.

rafmudaf (approved):

.clang-format
faisal-bhuiyan: I do want to have a line length limit here. Sorry for this small tweak, it was not necessary, I just like this number better. Do you want me to revert it?

ddement: No, it's fine. I figured doing line length manually would just save us from doing tweaks like this every few commits. We can keep the limit, if that works better for you.

faisal-bhuiyan: Thanks for your take on this, David. I do want to enforce a line limit since it can easily get out of control with super long lines. The default value for the Google style guide used in OpenTurbine is 80, which is kind of short. I just wanted to find something that's larger than 80 but not too large that it doesn't fit on standard monitors developers work on.

faisal-bhuiyan:
@ddement Not sure if you wanted to review this some more, if not, I can go ahead and merge this one.

Issue #19, opened 2023-03-29
Introduce quaternions and implement exponential mapping of quaternions to facilitate mapping from/to 3-D vectors


faisal-bhuiyan:
To enhance the capabilities to represent orientations and rotations in three-dimensional space, we should look into introducing the [quaternion number system](https://en.wikipedia.org/wiki/Quaternion) and implementing quaternion algebra, which extends the concept of complex numbers to four dimensions. Quaternion algebra involves defining operations such as addition, multiplication, conjugation, and normalization and has several advantages over other methods for representing rotations in three-dimensional space, including avoiding [gimbal lock](https://en.wikipedia.org/wiki/Gimbal_lock) and allowing for smooth interpolation between orientations.

**Breakdown of Tasks**

- _Implement quaternion addition, multiplication, conjugation, and normalization functions._

  To implement the above, we should introduce the concept of quaternions as a 4-D number system. To support quaternion algebra, we need to provide functions for quaternion addition, multiplication, conjugation, and normalization. 

- _Implement exponential mapping of quaternions to enable conversion from and to 3-D vectors._

  The quaternion exponential function, `exp(q)`, takes a quaternion `q` and computes its exponential in the quaternion space. It is defined as:
    
    `exp(q) = cos(||v||) + (v/||v||) * sin(||v||)`
    
    where `q = a + v`, `a` is the scalar part of the quaternion, `v` is the vector part of the quaternion, and `||v||` denotes the magnitude of the vector `v`.
    
    The quaternion exponential maps the Lie algebra of quaternions (i.e., the vector space of pure quaternions) to the corresponding Lie group, which in this case is the group of unit quaternions. The unit quaternions form a 3-D hypersphere embedded in 4-D space and represent rotations in three-dimensional space.

**Expected Outcome**

- A codebase that supports quaternions and quaternion algebra
- Methods for converting 4-D quaternions to 3-D physical vectors and vice-versa

PR #32, opened 2023-07-07
Introduce quaternions and implement exponential mapping

faisal-bhuiyan:
Implements #19.

### DIFF
--- a/src/rigid_pendulum_poc/CMakeLists.txt
+++ b/src/rigid_pendulum_poc/CMakeLists.txt
@@ -3,6 +3,7 @@ target_sources(${oturb_lib_name}
     #C++
     # diagnostics.cpp
     # io.cpp
+    quaternion.cpp
     solver.cpp
     # IOManager.cpp
 )
--- a/src/rigid_pendulum_poc/quaternion.cpp
+++ b/src/rigid_pendulum_poc/quaternion.cpp
@@ -1,7 +1,5 @@
 #include "src/rigid_pendulum_poc/quaternion.h"
 
-#include "src/utilities/log.h"
-
 namespace openturbine::rigid_pendulum {
 
 bool close_to(double a, double b) {
--- a/src/rigid_pendulum_poc/quaternion.cpp
+++ b/src/rigid_pendulum_poc/quaternion.cpp
@@ -1,0 +1,17 @@
+#include "src/rigid_pendulum_poc/quaternion.h"
+
+#include "src/utilities/log.h"
+
+namespace openturbine::rigid_pendulum {
+
+Quaternion::Quaternion(std::array<double, 4> values) : values_(std::move(values)) {
+}
+
+Quaternion::Quaternion(double p0, double p1, double p2, double p3) : values_({p0, p1, p2, p3}) {
+}
+
+Quaternion::Quaternion(double scalar, const std::array<double, 3>& vector)
+    : values_({scalar, vector[0], vector[1], vector[2]}) {
+}
+
+}  // namespace openturbine::rigid_pendulum
--- a/src/rigid_pendulum_poc/quaternion.cpp
+++ b/src/rigid_pendulum_poc/quaternion.cpp
@@ -19,14 +19,8 @@ bool close_to(double a, double b) {
     return (delta / a) < kTOLERANCE ? true : false;
 }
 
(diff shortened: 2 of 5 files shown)


michaelasprague:
@faisal-bhuiyan Good to see progress; haven't looked at code in detail, but two comments:
1.  I see you overloaded the plus and minus operators.  If you remember from our group meet, @ddement made strong case against overloading so that we can better facilitate performance portability.  Rather than overload the operators, can we simply make a data decision about where quaternions are located with respect to linear displacements in the generalized vector? 
2. I don't see a unit test; shouldn't that be the one of the earliest commits?  Did I miss them?

faisal-bhuiyan:
> 1. I see you overloaded the plus and minus operators.  If you remember from our group meet, @ddement made strong case against overloading so that we can better facilitate performance portability.  Rather than overload the operators, can we simply make a data decision about where quaternions are located with respect to linear displacements in the generalized vector?

I believe the overloading point was made for `Kokkos::Views` data structures. As things currently stand, quaternions are not based on `Kokkos` but `std::array`. Feel free to chime in, @ddement.

> 2. I don't see a unit test; shouldn't that be the one of the earliest commits?  Did I miss them?

Unit tests were part of almost every commit, as is typical with TDD. The file could be found here: https://github.com/Exawind/openturbine/blob/7db8f91398d0e248e221f05f07ff2ad5d06a1b00/tests/unit_tests/rigid_pendulum_poc/test_quaternions.cpp

ddement:
> I believe the overloading point was made for `Kokkos::Views` data structures. As things currently stand, quaternions are not based on `Kokkos` but `std::array`.

We don't want to use std::array for a different reason.  We can't promise that the standard library we're using will be GPU compatible, so it's best to avoid it for anything that we want to run on device.  For this quaternion type, we can just have the four doubles explicitly be members of the class.  This, along with some function decorations, should let work with the GPU.  

For performance, it really depends on usage.  I was mainly referring to larger calls which would launch a kernel on the GPU.  For small calls that would be used from inside a kernel (i.e. inside the parallel for) and using only variables defined on the stack, the performance loss will be much less when compared to a hand written calculation. 

Do you plan to make this the main data type?  Will we be seeing something like `Kokkos::View<Quaternion*>`, or will you just be populating it right before using the variables for calculation?  Off hand, I prefer the second option.

faisal-bhuiyan:
@ddement 
> We don't want to use std::array for a different reason. We can't promise that the standard library we're using will be GPU compatible, so it's best to avoid it for anything that we want to run on device. For this quaternion type, we can just have the four doubles explicitly be members of the class. This, along with some function decorations, should let work with the GPU.


Converted to the above with the latest commit. Let's move forward with reviewing the rest of the PR. 


> Do you plan to make this the main data type? Will we be seeing something like `Kokkos::View<Quaternion*>`, or will you just be populating it right before using the variables for calculation? Off hand, I prefer the second option.

I was indeed planning to do the `Kokkos::View<Quaternion*>` - looks like it's not the optimum choice? Could you expand on the second option a little more? Thanks!

ddement:
> I was indeed planning to do the `Kokkos::View<Quaternion*>` - looks like it's not the optimum choice? Could you expand on the second option a little more? Thanks!

For the second option, there are a few choices.  For the simplest version, our main data structure would be a `Kokkos::View<double*>` which contains the quaternion data.  It gives us lots of options with regard to data ordering, which we can work through in the future, while the "view of quaternions" locks us into a single ordering.  To use your new quaternion type, we'd create them locally within the parallel_for loop.  It would look something like:

```
Kokkos::View<double*> data = ...;
Kokkos::View<double*> output= ...;
Kokkos::parallel_for(execution_policy, KOKKOS_LAMBDA (int i) {
  ...
  Quaternion q {data[i+0], data[i+1], data[i+2], data[i+3]}; //populate quaternion with our data
  ...
  Quaternion outputQ = do_math(q); //do the desired computation
  output[i+0] = outputQ.firstComponent();
  output[i+1] = outputQ.seciondComponent();
  ...
});
```

We can always proceed the way you had initially planned and then refactor once we have a working prototype.  The project is still early enough that it will not be a significant time cost to do so, and it may help both of us to have a concrete example to work from.

ddement:
The class itself (and the testing) look great to me.

alanw0 (approved):
Looks good to me.
I like the tests, they will make it easy to keep things correct if we refactor things when getting it working on GPU etc.

PR #36, opened 2023-07-11
Resolve missing lapack/blas symbols.

alanw0:
On my linux system, I loaded a lapack module, which includes lapacke and lapack and blas.
But linking openturbine failed with unresolved lapack and blas symbols until I explicitly added lapack and blas to the CMakeLists.txt file which was already referencing lapacke.

Apparently this isn't needed on all systems, but perhaps it is good to explicitly reference the lapack and blas libraries?

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -8,6 +8,8 @@ target_link_libraries(${oturb_exe_name} PRIVATE
     Kokkos::kokkos
     LAPACK::LAPACK
     lapacke
+    lapack
+    blas
 )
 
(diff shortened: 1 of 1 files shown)


faisal-bhuiyan (approved):
> Apparently this isn't needed on all systems, but perhaps it is good to explicitly reference the lapack and blas libraries?

I agree, let's add them to the list.

ddement (approved):
Builds on Sandia's ASCICGPU057 machine with these changes.  Being more specific doesn't hurt anything.

Issue #38, opened 2023-07-12
Introduce quaternion-based representation of 3-D rotations in OpenTurbine


faisal-bhuiyan:
Building upon the [existing concept and implementation of quaternions and quaternion algebra](https://github.com/Exawind/openturbine/pull/32), we should now focus on implementing 3D rotations using quaternions. Quaternion algebra offers numerous advantages over traditional rotation representation methods, such as avoiding [gimbal lock](https://en.wikipedia.org/wiki/Gimbal_lock) and enabling smooth interpolation between orientations. The provided mathematical terms in the [linked document](https://www.overleaf.com/read/jrxxvrckpspb) will serve as a reference for the implementation and usage of quaternions in OpenTurbine.

PR #39, opened 2023-07-14
Implement rotation of 3-D vectors using quaternions

faisal-bhuiyan:
This PR introduces a method for rotating physical vectors using quaternions, implementing #38. The implementation includes the following key features:

- Introduces a new method for rotating vectors using quaternions.
- Adds helper methods for converting between axis-angle representation and quaternions.
- Utilizes test-driven development to ensure comprehensive testing of the implemented features.

### DIFF
--- a/tests/unit_tests/rigid_pendulum_poc/test_math_utilities.cpp
+++ b/tests/unit_tests/rigid_pendulum_poc/test_math_utilities.cpp
@@ -4,76 +4,260 @@
 
 namespace openturbine::rigid_pendulum::tests {
 
-TEST(MathUtilitiesTest, CloseTo) {
+TEST(MathUtilitiesTest, CloseTo_Set1) {
     ASSERT_TRUE(close_to(1., 1.));
+}
+
+TEST(MathUtilitiesTest, CloseTo_Set2) {
     ASSERT_TRUE(close_to(1., 1. + 1e-7));
+}
+
+TEST(MathUtilitiesTest, CloseTo_Set3) {
     ASSERT_TRUE(close_to(1., 1. - 1e-7));
+}
+
+TEST(MathUtilitiesTest, CloseTo_Set4) {
     ASSERT_FALSE(close_to(1., 1. + 1e-5));
+}
+
+TEST(MathUtilitiesTest, CloseTo_Set5) {
     ASSERT_FALSE(close_to(1., 1. - 1e-5));
+}
+
+TEST(MathUtilitiesTest, CloseTo_Set6) {
     ASSERT_TRUE(close_to(1e-7, 1e-7));
+}
 
+TEST(MathUtilitiesTest, CloseTo_Set7) {
     ASSERT_TRUE(close_to(-1., -1.));
-    ASSERT_TRUE(close_to(-1., -1. + 1e-7));
-    ASSERT_TRUE(close_to(-1., -1. - 1e-7));
-    ASSERT_FALSE(close_to(-1., -1. + 1e-5));
-    ASSERT_FALSE(close_to(-1., -1. - 1e-5));
-    ASSERT_TRUE(close_to(-1e-7, -1e-7));
+}
 
+TEST(MathUtilitiesTest, CloseTo_Set8) {
+    ASSERT_TRUE(close_to(-1., -1. + 1e-7));
+}
+
+TEST(MathUtilitiesTest, CloseTo_Set9) {
+    ASSERT_TRUE(close_to(-1., -1. - 1e-7));
+}
+
+TEST(MathUtilitiesTest, CloseTo_Set10) {
+    ASSERT_FALSE(close_to(-1., -1. + 1e-5));
+}
+
+TEST(MathUtilitiesTest, CloseTo_Set11) {
+    ASSERT_FALSE(close_to(-1., -1. - 1e-5));
+}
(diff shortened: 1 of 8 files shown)


tests/unit_tests/rigid_pendulum_poc/test_quaternions.cpp
```diff
@@ -236,4 +236,236 @@ TEST(QuaternionTest, GetRotationVectorFromNullQuaternion) {
     ASSERT_NEAR(std::get<2>(v), std::get<2>(expected), 1e-6);
 }
 
+TEST(QuaternionTest, WrapAngleToPi) {
```
ddement, line 239: This should be many different tests (one for each angle).  If we ever refactor wrap_angle_to_pi and introduce a bug, this will make debugging a little easier.

faisal-bhuiyan: image

Just wondering if we need more specific information than this for debugging - it provides the exact line # where the assertion fails, along with the calculated and expected values. My opinion is to keep these helper function testing condensed if we can help it, but open to your thoughts.

ddement: It does. But "assert" macros end the test at that failure, so we don't know if there were any other failures later in the test.  

Also, long tests are harder to understand as a new person reading them, and it's hard to tell at first glance that each grouping is fully independent.

Making each independent condition its own test is the simplest way to solve both of these problems

faisal-bhuiyan: Sounds reasonable to me, let's break them up.

ddement (approved):

tests/unit_tests/rigid_pendulum_poc/test_math_utilities.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+namespace openturbine::rigid_pendulum::tests {
+
+TEST(MathUtilitiesTest, CloseTo) {
+    ASSERT_TRUE(close_to(1., 1.));
+    ASSERT_TRUE(close_to(1., 1. + 1e-7));
+    ASSERT_TRUE(close_to(1., 1. - 1e-7));
+    ASSERT_FALSE(close_to(1., 1. + 1e-5));
+    ASSERT_FALSE(close_to(1., 1. - 1e-5));
+    ASSERT_TRUE(close_to(1e-7, 1e-7));
```
alanw0, line 28: Why do you use ASSERT instead of EXPECT?
I don't think it matters very much, but ASSERT will stop the program right there, while EXPECT will let it continue so you can see how many failures there are, instead of just the first failure.

tests/unit_tests/rigid_pendulum_poc/test_math_utilities.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include "src/rigid_pendulum_poc/utilities.h"
+
+namespace openturbine::rigid_pendulum::tests {
+
+TEST(MathUtilitiesTest, CloseTo) {
+    ASSERT_TRUE(close_to(1., 1.));
+    ASSERT_TRUE(close_to(1., 1. + 1e-7));
+    ASSERT_TRUE(close_to(1., 1. - 1e-7));
+    ASSERT_FALSE(close_to(1., 1. + 1e-5));
+    ASSERT_FALSE(close_to(1., 1. - 1e-5));
```
alanw0, line 24: It's good to test close_to with values around the tolerance.
Perhaps it would be more clear to use the actual tolerance constant, with `kTol*10` and `kTol/10`.
Will tolerance always be 1e-6? Could some problems benefit from a smaller or larger tolerance? Should it be related to edge size in the mesh, or some other problem characteristic? If it should just always be 1e-6 then these tests are fine as is.

alanw0 (approved):

tests/unit_tests/rigid_pendulum_poc/test_math_utilities.cpp
faisal-bhuiyan: Thanks for the helpful tip, @alanw0 - will use `EXPECT` statements going forward where it does not invalidate the rest of the test. 

For the above set,  I changed the tests to one `ASSERT_TRUE` statement per test per @ddement's suggestion. Let's keep them unchanged.

tests/unit_tests/rigid_pendulum_poc/test_math_utilities.cpp
faisal-bhuiyan: > It's good to test close_to with values around the tolerance.
> Perhaps it would be more clear to use the actual tolerance constant, with `kTol*10` and `kTol/10`.

Yup, I agree. Thanks for the suggestion.

> Will tolerance always be 1e-6? Could some problems benefit from a smaller or larger tolerance? Should it be related to edge size in the mesh, or some other problem characteristic? If it should just always be 1e-6 then these tests are fine as is.

Just want to point out the tolerance is not static, only the default is 1e-6. We should be able to change the value based on the problem at hand like you suggested above.

https://github.com/Exawind/openturbine/blob/a3d068c26222fa1c75c249a38de4577acab9af8e/src/rigid_pendulum_poc/utilities.h#L16

Issue #17, opened 2023-03-28
Establish the geometric concept of Vectors


faisal-bhuiyan:
As part of the common mathematical library `oturb::common::geom`, we should introduce the concept of a `Vector` class based on the basis vectors of the three dimensional Cartesian coordinate system. The specific requirements of such a class can be defined as follows:
- Provide appropriate constructors to facilitate the instantiation of vectors in different ways
- Implement operations such as 
  - determining the inner and cross products between two vector entities (likely via overloaded operators)
  - calculating length of the vector
  - finding the unit vector in the direction of a provided vector etc.
- Define some useful checks such as
  -  if two provided vectors are orthogonal to each other
  -  if two vectors are parallel to each other etc.

As always, the implemented logic should be complemented with unit tests having comprehensive code coverage (or as close as possible to it), which should provide as validation/verification of the work and provide a good "document" of examples on how to use the class/functions for future developers.

PR #40, opened 2023-07-17
Introduce the concept of 3-D vectors in OpenTurbine

faisal-bhuiyan:
Implements #17.

This pull request introduces the `Vector` class representing a three-dimensional vector based on the basis vectors of the Cartesian coordinate system. The class provides the following functionalities:

- Constructors to instantiate vectors in different ways.
- Overloaded operators for determining the inner product and cross product between two vector entities.
- Calculation of the length of the vector.
- Calculation of the unit vector in the direction of a provided vector.
- Checks for orthogonality and parallelism between two vectors.

Comprehensive unit tests have been implemented to ensure code coverage and validate the functionality of the `Vector` class. These tests serve as examples for future developers on how to use the class and its functions effectively.

### DIFF
--- a/src/rigid_pendulum_poc/CMakeLists.txt
+++ b/src/rigid_pendulum_poc/CMakeLists.txt
@@ -9,5 +9,6 @@ target_sources(${oturb_lib_name}
     state.cpp
     time_integrator.cpp
     time_stepper.cpp
+    vector.cpp
     # IOManager.cpp
 )
--- a/src/rigid_pendulum_poc/quaternion.cpp
+++ b/src/rigid_pendulum_poc/quaternion.cpp
@@ -27,7 +27,7 @@ Quaternion Quaternion::GetUnitQuaternion() const {
 }
 
 Quaternion quaternion_from_rotation_vector(const Vector& vector) {
-    auto [v0, v1, v2] = vector;
+    auto [v0, v1, v2] = vector.GetComponents();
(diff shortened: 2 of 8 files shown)


ddement (approved):
Looks good

alanw0 (approved):
I think this looks generally good, I'm happy to see the unit-testing as always.
However, it seems unfortunate that we need to write this ourselves from scratch. There is a vector class that already exists in nalu-wind, which I believe came from something in amr-wind... It has some good features, such as it's already GPU-friendly, etc. Perhaps it would be worth considering to grab a copy of it, and then add missing features that we need. You can see it here:
https://github.com/Exawind/nalu-wind/blob/master/include/vs/vector.h

But your commit looks fine, I wouldn't object if you go ahead and merge.

faisal-bhuiyan:
> I think this looks generally good, I'm happy to see the unit-testing as always. However, it seems unfortunate that we need to write this ourselves from scratch. There is a vector class that already exists in nalu-wind, which I believe came from something in amr-wind... It has some good features, such as it's already GPU-friendly, etc. Perhaps it would be worth considering to grab a copy of it, and then add missing features that we need. You can see it here: https://github.com/Exawind/nalu-wind/blob/master/include/vs/vector.h

Thanks for the comment and suggestion, @alanw0. Please feel free to open an issue for any improvements you think might be worthwhile to implement in OpenTurbine.

> But your commit looks fine, I wouldn't object if you go ahead and merge.

I am going forward with merging this one.

Issue #43, opened 2023-07-21
Implement Lie group time integrator algorithm as an extension of the classical Generalized-Alpha integrator


faisal-bhuiyan:
The goal of this issue is to implement the Lie group time integrator algorithm as an extension of the existing classical generalized-alpha integrator. This algorithm, as proposed in the paper ["On the Use of Lie Group Time Integrators in Multibody Dynamics"](https://asmedigitalcollection.asme.org/computationalnonlinear/article-abstract/5/3/031002/384896/On-the-Use-of-Lie-Group-Time-Integrators-in?redirectedFrom=fulltext), offers a novel solution to the rotation parametrization problem and allows for the simulation of flexible multibody systems with higher accuracy and efficiency. In our work, we will use unit quaternions to represent 3-D rotations, which form a lie group.

![image](https://github.com/Exawind/openturbine/assets/59463700/141ba40a-5c42-4a40-b8b2-88b67f211a3b)

PR #44, opened 2023-07-24
Lie group generalized alpha implementation

faisal-bhuiyan:
Implements #43.

Important additions in this work:
- I**mplementation of a POC time integrator, including nonlinear system solver**: The classical generalized-alpha time integrator implemented previously was updated to a Lie group based time integrator in this work. Follows the algorithm from Table 1 of [Brüls, Cardona, Arnold](https://doi.org/10.1016/j.mechmachtheory.2011.07.017).
- **Verification/demonstration of the solution for a rigid-body-dynamics problem**: As a suitable test problem, the heavy top problem was implemented to provide a concrete test case for the residual and iteration matrices. Follows the theoretical work from [Brüls and Cardona (2010)](https://doi.org/10.1115/1.4001370).


src/rigid_pendulum_poc/generalized_alpha_time_integrator.cpp
```diff
@@ -317,6 +317,42 @@ HostView2D heavy_top_tangent_damping_matrix(
...
+            for (size_t j = 0; j < 6; j++) {
+                if (i < 3 && j < 3) {
+                    tangent_stiffness_matrix(i, j) = 0.;
+                } else if (i < 3 && j >= 3) {
+                    tangent_stiffness_matrix(i, j) = 0.;
+                } else if (i >= 3 && j < 3) {
+                    tangent_stiffness_matrix(i, j) = 0.;
+                } else {
+                    tangent_stiffness_matrix(i, j) = non_zero_block(i - 3, j - 3);
+                }
```
alanw0: Perhaps this logic could be more clear if written like this?
```
if (i >= 3 && j >= 3) {
  tangent_stiffness_matrix(i, j) = non_zero_block(i - 3, j - 3);
}
else {
  tangent_stiffness_matrix(i, j) = 0.;
}
```

src/rigid_pendulum_poc/generalized_alpha_time_integrator.cpp
```diff
@@ -106,111 +109,125 @@ std::tuple<State, HostView1D> GeneralizedAlphaTimeIntegrator::AlphaStep(
...
+        // Compute the iteration matrix and solve the linear system to get the increments
         auto iteration_matrix = ComputeIterationMatrix(gen_coords, matrix);
-        auto gen_coords_delta = residuals;
-        auto constraints_delta = constraints;
-        solve_linear_system(iteration_matrix, gen_coords_delta);
-        solve_linear_system(iteration_matrix, constraints_delta);
+        auto delta_x = residuals;
+        auto delta_constraints = constraints;
 
+        solve_linear_system(iteration_matrix, delta_x);
```
michaelasprague, line 126: This must be a single-system solve -- the iteration matrix is not block diagonal.  See Eq. (34) in Bruls et al 2012

src/rigid_pendulum_poc/generalized_alpha_time_integrator.cpp
```diff
@@ -106,111 +109,125 @@ std::tuple<State, HostView1D> GeneralizedAlphaTimeIntegrator::AlphaStep(
...
+
+        // Update the velocity, acceleration, and constraints based on the increments
         Kokkos::parallel_for(
             size,
             KOKKOS_LAMBDA(const int i) {
-                gen_coords(i) += gen_coords_delta(i);
-                gen_velocity(i) += gamma_prime * gen_coords_delta(i);
-                gen_accln(i) += beta_prime * gen_coords_delta(i);
-                constraints(i) += constraints_delta(i);
+                x(i) += delta_x(i);
```
michaelasprague, line 133: See Table 1 in Bruls et al. (2012); should be delta_x / h

src/rigid_pendulum_poc/generalized_alpha_time_integrator.cpp
```diff
@@ -106,111 +109,125 @@ std::tuple<State, HostView1D> GeneralizedAlphaTimeIntegrator::AlphaStep(
...
+                velocity(i) += GAMMA_PRIME * delta_x(i);
+                acceleration(i) += BETA_PRIME * delta_x(i);
+                constraints(i) += delta_constraints(i);
```
michaelasprague, line 136: I don't think we want to call this variable constraints; maybe lagmult or something short for LagrangeMultiplier.

src/rigid_pendulum_poc/generalized_alpha_time_integrator.cpp
faisal-bhuiyan: Thanks for catching this! Corrected.

faisal-bhuiyan:
@michaelasprague @alanw0 @ddement: This pull request introduces the Lie group based time integrator, which is now fully implemented and tested. While there may be some minor cleanup required, it's ready for review to gather feedback and address any potential improvements. Please feel free to ask questions on the implementation and the relevant theory.

tests/unit_tests/rigid_pendulum_poc/test_heavy_top.cpp
```diff
@@ -6,6 +6,37 @@
 
 namespace openturbine::rigid_pendulum::tests {
 
+TEST(HeavyTopProblemFromBrulsAndCardona2010PaperTest, CalculateResidualVector) {
+    auto M = MassMatrix(15., Vector(0.234375, 0.46875, 0.234375));
+    auto mass_matrix = M.GetMassMatrix();
+    auto rotation_matrix = create_matrix(
+        {{0.617251, -0.757955, 0.210962},
```
alanw0, line 13: I'm usually not enthusiastic about comments, but in this case it might be nice to put a note that references the paper or source for these hard-coded constants.

src/rigid_pendulum_poc/heavy_top.cpp
```diff
@@ -1,88 +1,44 @@
...
-                    tangent_stiffness_matrix(i, j) = non_zero_block(i - 3, j - 3);
-                }
-            }
-        }
-    );
+    auto log = util::Log::Get();
+    log->Debug("Residual vector is " + std::to_string(6) + " x 1 with elements\n");
+    for (size_t i = 0; i < 6; i++) {
+        log->Debug(std::to_string(residual_vector(i)) + "\n");
+    }
```
alanw0, line 76: I think this looks good in general. We may need to make some adjustments (probably just add a deep-copy to bring residual_vector back to host) once we start getting things to run on GPU. Maybe we would only do that for debug-mode, if logging is only active in debug-mode, or something like that.

alanw0 (approved):

alanw0 (approved):

src/rigid_pendulum_poc/heavy_top.cpp
faisal-bhuiyan: @alanw0  I am running into a peculiar memory related issue as follows. I wonder if it has something to do with copying Kokkos::Views as you describe above. Do you have any suggestions as to how to navigate this or expand on the above?

```
openturbine_unit_tests(43306,0x1ff149e00) malloc: Heap corruption detected, free list is damaged at 0x600003644620
*** Incorrect guard value: 0
openturbine_unit_tests(43306,0x1ff149e00) malloc: *** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```

alanw0: If you are not running on GPU, you shouldn't run into the issue I was anticipating above.
Assuming you are running in a pure CPU build, you'll probably need to run this through valgrind or asan to see where the memory error is coming from. If you don't have access to one of those tools, we can do that for you and help figure it out.

faisal-bhuiyan: I am running on CPU only and not completely sure about the source of the error here. It would be really helpful if you could take a look and I will also keep investigating myself - this appears to be the only remaining hurdle in completing the heavy top problem.

I am pushing the remaining local changes from my end - please try to run the `TEST(HeavyTopProblemFromBrulsAndCardona2010PaperTest, AlphaStepSolutionAfterOneInc)` unit test. It appears fpr this particular test with high value for the `double final_time{0.2}` variable.

alanw0: ok I'll see what I can figure out.

alanw0: valgrind says there is an 'invalid write', which usually means writing past the end of an array allocation, at line 108 of `generalized_alpha_time_integrator.cpp`. That line is `lagrange_mults_next(i) = 0.;`. It is in a Kokkos::parallel_for which is looping over `velocity.size()`. So it is assuming that velocity and lagrange_mults_next are the same length, and they actually are different lengths?

faisal-bhuiyan: They are not the same size, this was exactly what the issue was. Thanks for your help, greatly appreciated!

alanw0: valgrind also says there is a bad read at line 164 of utilities.cpp, which is `sum += matrix_a(row, i) * matrix_b(i, column);`.
Again this hints at an array-bounds problem, but it isn't obvious to me what the problem is. It looks like you've got the sizes correct, and you've made sure that n_rows == n_columns, so it seems like it should be ok...

faisal-bhuiyan: Ok I will look into it.

alanw0: I think I see it now, on line 158/159 where you set the 'entries' object, it should have bounds of A_n_rows by B_n_columns.

faisal-bhuiyan: Yup, I see that as well.

alanw0:
One last thing flagged by valgrind: generalized_alpha_time_integrator.cpp line 297 is another loop that accesses lagrange_mults_next with loop-bounds of velocity.size().

ddement (commented):
We'll need to do some refactoring once we get a solution to the milestone problem (that way we can rely on the tests to ensure we haven't broken anything).  

In particular, we'll need to start to pay attention to memory/execution spaces and function decorations much more carefully.

The `AlphaStep` function is quite a bit too long, in my opinion.  We'll want to find a way to refactor it to make it simpler and easier to read.  

Our function interfaces are also rather long (accepting a bunch of arguments).  My guess is that we can fix this through the use of member variables.  We're about at the point that we can have a discussion about what data structures.  

There are also several places that you do a fairly small amount of work in a parallel_for loop.  Depending on how it scales with problem size (I know our current problems are tiny), it may be better to just do that work on host and copy it to device.  We can address this when we start cleaning up our Kokkos usage.

alanw0 (approved):
I think this is pretty much ready to merge, subject to possibly addressing a some of David's comments. Faisal, use your judgment on addressing those now, or in a follow-up PR, since at least many of them are related to GPU readiness.

I'll propose a couple of things that can also be done in follow-up PRs if you wish:
1. The generalized-alpha file should be general, ideally it shouldn't need to refer to specific problem types such as heavy-top etc. This can be fixed by passing in a function for iteration-matrix etc., as David noted. More generally, I might even prefer having a base-class called "Problem" or something similar, and it could have methods like 'get_iteration_matrix' etc. Then the generalized-alpha code doesn't need to know what kind of problem is being solved. At the calling scope, e.g. the unit-test, you create a specialization like 'HeavyTopProblem' and pass that in.
2. I agree that the AlphaStep function is quite long. However, a lot of that code is the logging stuff. Do you intend that the logging code should be in there permanently, or do you intend to delete it once the math is settled down and well tested? I think logging code is a debugging mechanism, which shouldn't be needed as much if testing is adequate.

Anyway, overall, I think the PR is basically ready to merge.

src/rigid_pendulum_poc/generalized_alpha_time_integrator.cpp
```diff
@@ -55,163 +38,431 @@ GeneralizedAlphaTimeIntegrator::GeneralizedAlphaTimeIntegrator(
...
-    // TODO: Provide actual constraints
-    auto constraints = HostView1D("constraints", size);
+            delta_gen_coords(i) = velocity(i) + h * (0.5 - kBETA_) * algo_acceleration(i) +
+                                  h * kBETA_ * algo_acceleration_next(i);
+            velocity(i) +=
+                h * (1 - kGAMMA_) * algo_acceleration(i) + h * kGAMMA_ * algo_acceleration_next(i);
+
+            algo_acceleration(i) = algo_acceleration_next(i);
+
+            acceleration(i) = 0.;
```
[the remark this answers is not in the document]
faisal-bhuiyan: Yup, this is what the algorithm does at this stage.

Screenshot 2023-08-02 at 3 59 50 PM

src/rigid_pendulum_poc/generalized_alpha_time_integrator.cpp
```diff
@@ -55,163 +38,431 @@ GeneralizedAlphaTimeIntegrator::GeneralizedAlphaTimeIntegrator(
...
+        r.GetYComponent(),       // component 2
+        r.GetZComponent(),       // component 3
+        q.GetScalarComponent(),  // component 4
+        q.GetXComponent(),       // component 5
+        q.GetYComponent(),       // component 6
+        q.GetZComponent()        // component 7
+    };
+
+    Kokkos::parallel_for(
+        gen_coords.size(), KOKKOS_LAMBDA(const size_t i) { gen_coords_next(i) = components[i]; }
```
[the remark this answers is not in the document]
faisal-bhuiyan: Sounds like a good plan to me. At this stage, we do have a functional CPU implementation. Maybe work on this with a well-defined and scoped issue such as "Ensure GPU compatibility in Generalized Alpha calculations" and submit a PR?

src/rigid_pendulum_poc/generalized_alpha_time_integrator.cpp
```diff
@@ -55,163 +38,431 @@ GeneralizedAlphaTimeIntegrator::GeneralizedAlphaTimeIntegrator(
...
+    auto algo_acceleration = HostView1D("algorithmic_acceleration", acceleration.size());
+    Kokkos::deep_copy(algo_acceleration, state.GetAlgorithmicAcceleration());
+
+    // Initialize some X_next variables to assist in updating the State - only for ones that
+    // require both the current and next values
+    auto gen_coords_next = HostView1D("gen_coords_next", gen_coords.size());
+    auto algo_acceleration_next =
+        HostView1D("algorithmic_acceleration_next", algo_acceleration.size());
+    auto delta_gen_coords = HostView1D("gen_coords_increment", velocity.size());
+    auto lagrange_mults_next = HostView1D("lagrange_mults_next", lagrange_mults.size());
```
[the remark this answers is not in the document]
faisal-bhuiyan: I agree with the above. Let's put a pin on this and I'll address this on a follow up PR.

michaelasprague:
@faisal-bhuiyan @rafmudaf -- Can we create a "heavy top" regression test?

ddement (approved):

faisal-bhuiyan:
> I think this is pretty much ready to merge, subject to possibly addressing a some of David's comments. Faisal, use your judgment on addressing those now, or in a follow-up PR, since at least many of them are related to GPU readiness.
> 
> I'll propose a couple of things that can also be done in follow-up PRs if you wish:
> 
> 1. The generalized-alpha file should be general, ideally it shouldn't need to refer to specific problem types such as heavy-top etc. This can be fixed by passing in a function for iteration-matrix etc., as David noted. More generally, I might even prefer having a base-class called "Problem" or something similar, and it could have methods like 'get_iteration_matrix' etc. Then the generalized-alpha code doesn't need to know what kind of problem is being solved. At the calling scope, e.g. the unit-test, you create a specialization like 'HeavyTopProblem' and pass that in.
> 2. I agree that the AlphaStep function is quite long. However, a lot of that code is the logging stuff. Do you intend that the logging code should be in there permanently, or do you intend to delete it once the math is settled down and well tested? I think logging code is a debugging mechanism, which shouldn't be needed as much if testing is adequate.
> 
> Anyway, overall, I think the PR is basically ready to merge.

Thanks for the review, @alanw0. Above suggestions make sense to and I implemented both of them in the recent commits. Please take a one last look to make sure things are looking better.

tests/unit_tests/rigid_pendulum_poc/test_heavy_top.cpp
```diff
@@ -135,37 +135,45 @@ TEST(HeavyTopProblemFromBrulsAndCardona2010PaperTest, CalculateConstraintGradien
...
-            {0., 0., 0., -0.768003, 1.404458, -0.046875, 0.210962, -0.004165, 0.977485},  // row 4
-            {0., 0., 0., 0., 0.46875, 0., 0., 0., 0.},                                    // row 5
-            {0., 0., 0., 0.046875, 3.275712, -0.768003, -0.617251, -0.775967, 0.129909},  // row 6
-            {-1., 0., 0., 0.210962, 0., -0.617251, 0., 0., 0.},                           // row 7
-            {0., -1., 0., -0.004165, 0., -0.775967, 0., 0., 0.},                          // row 8
-            {0., 0., -1., 0.977485, 0., 0.129909, 0., 0., 0.}                             // row 9
+            {15., 0., 0., 0., 0., 0., -1., 0., 0.},                    // row 1
+            {0., 15., 0., 0., 0., 0., 0., -1., 0.},                    // row 2
+            {0., 0., 15., 0., 0., 0., 0., 0., -1.},                    // row 3
+            {0., 0., 0., -1.765625, 0.625000, -0.046875, 0., 0., 1.},  // row 4
```
alanw0, line 185: Why did these numbers change?

alanw0 (approved):

tests/unit_tests/rigid_pendulum_poc/test_heavy_top.cpp
faisal-bhuiyan: I modified the function signature and had to redo the tests - values are verified against the prototype Fortran code after another round of revision.

alanw0 (approved):

PR #48, opened 2023-08-07
Remove heat-solve code and also enable building on GPU

alanw0:
This pull request has two commits.
1. Remove the heat-solve code.
2. Make code updates in the alpha-solver to enable building on GPU.

The tests still don't run successfully on GPU yet, that will be done in follow-up work.


ddement (approved):
Looks good.  I can get to work making it actually run.

rafmudaf:
On remove the heat solver code, that looks good to me.

I noticed the `*_local` variables and I assume those are for the GPU device. If there are other nuances or implementation details that are particular to GPU capabilities, it would be good to create a corresponding section in the [developer docs](https://exawind.github.io/openturbine/dev_guide.html) to make note of these. Even stating something like "All device variables should use the convention "[hostvariable]_local"'. @alanw0 your call if it's relevant to add here or wait until there is more information to add to such a section, but it's something to keep in mind for the ongoing GPU-support.

rafmudaf (approved):

alanw0:
> On remove the heat solver code, that looks good to me.
> 
> I noticed the `*_local` variables and I assume those are for the GPU device. 

This is just for the specific case of capturing class-member variables into device kernels. It causes a cuda warning 'implicit this capture'. The solution is to first assign them to local-scope variables and then the capture doesn't produce a warning.

PR #49, opened 2023-08-11
Minor fixes

faisal-bhuiyan:
Some minor refactoring, renaming, and housekeeping done in this PR - no new feature was implemented in this work. 

Summary:

- Namespace changed from `rigid_pendulum` -> `gen_alpha_solver`
- Subdirectory name changed from `rigid_pendulum_poc` -> `gen_alpha_solver_poc`
- Update `const` and `Enum` names from ALLCAPS -> PascalCase as recommended in Google style guide
- Add parameter names to function declarations in header files where the parameters are not clear from the context, particularly for matrices and vectors
- Use named `const`s in place of "magic numbers" in source files  
- Consolidate unit tests using test fixture where appropriate and merge similar looking tests
- Fix a bug in the output `std::vector<State>` - it was surprisingly modifying the older (than the most recent increment) results


alanw0 (commented):
I'm curious why you want to define HostView1D and HostView2D in 4 or 5 different places, instead of just 1 place?
In what circumstance would you want them to have different definitions in some places?

Also, I thought we discussed waiting on major file/folder moves until after David makes the GPU changes? Conflicts are harder to resolve in files that have been moved or renamed.

faisal-bhuiyan:
> I'm curious why you want to define HostView1D and HostView2D in 4 or 5 different places, instead of just 1 place? In what circumstance would you want them to have different definitions in some places?

These aliases can be different for different places (obviously not here), and probably easier to find their definition in the header file rather than trying to hunt for it in a single location. What seems to be the standard practice?

> Also, I thought we discussed waiting on major file/folder moves until after David makes the GPU changes? Conflicts are harder to resolve in files that have been moved or renamed.

I do remember that conversation and we can sit on merging this PR until after the GPU related work gets merged. Does this sound reasonable?

alanw0:
If the aliases might have different definitions in different headers, then I would give them different names to avoid confusion. If they need to be passed from one scope to another, obviously they need to be the same type in both scopes. If they are the same, and need to be compatible for passing back and forth, then there should only be one definition since duplication causes higher maintenance effort and is error prone.

Yes let's hold off on merging until we see the extent of David's GPU changes. Great, thanks.

faisal-bhuiyan:
> If the aliases might have different definitions in different headers, then I would give them different names to avoid confusion. If they need to be passed from one scope to another, obviously they need to be the same type in both scopes. If they are the same, and need to be compatible for passing back and forth, then there should only be one definition since duplication causes higher maintenance effort and is error prone.

@alanw0  What do you think about the following design?  We define the aliases in a separate source file `openturbine_types.h` in the `openturbine::util` namespace as follows.

```
#pragma once

#include <Kokkos_Core.hpp>

namespace openturbine::util {

// Used to store a 1D Kokkos View of doubles on the host
using HostView1D = Kokkos::View<double*, Kokkos::HostSpace>;

// Used to store a 2D Kokkos View of doubles on the host
using HostView2D = Kokkos::View<double**, Kokkos::HostSpace>;

}  // namespace openturbine::util
```

When we need to use the above data structures, we bring that into the current scope.

```
using openturbine::util::HostView1D;
using openturbine::util::HostView2D;
```

alanw0:
> @alanw0 What do you think about the following design? We define the aliases in a separate source file `openturbine_types.h` in the `openturbine::util` namespace as follows.
> When we need to use the above data structures, we bring them into the current scope.

Yes that seems good.

ddement (approved):

faisal-bhuiyan:
> I think these changes look ok, but it's hard to be certain without running them in a cuda build. Have you been able to do a cuda build and run tests?

I have not done the cuda build - I'll make sure things look good there before merging this. I wonder if there's a good way to automate this via the CI pipeline.

ddement:
test HeavyTopProblemFromBrulsAndCardona2010PaperTest.CalculateTangentDampingMatrix aborts on CUDA builds with error:

Kokkos::View ERROR: attempt to access inaccessible memory space (label="nonzero_block").

faisal-bhuiyan:
> test HeavyTopProblemFromBrulsAndCardona2010PaperTest.CalculateTangentDampingMatrix aborts on CUDA builds with error:
> 
> Kokkos::View ERROR: attempt to access inaccessible memory space (label="nonzero_block").

@ddement  Thanks for pointing out the above. I currently don't have access to the NREL supercomputer (it should be resolved in a few days) to go through the CUDA build myself, could you help me debug this issue? Specifically, I looked into the changes made to the `HeavyTopLinearizationParameters::TangentDampingMatrix` and reverted a change with the latest commit. Please try to build and run the unit tests on CUDA.

ddement:
@faisal-bhuiyan I fixed the test failures.  There were a couple of places where you accidentally reverted DefaultExecutionSpace to DefaultHostExecutionSpace.  Everything should be good to push

faisal-bhuiyan:
> @faisal-bhuiyan I fixed the test failures. There were a couple of places where you accidentally reverted DefaultExecutionSpace to DefaultHostExecutionSpace. Everything should be good to push

Thanks for taking a look at this and fixing it, @ddement.

Issue #16, opened 2023-03-28
Introduce the concept of points in OpenTurbine


faisal-bhuiyan:
We need to establish a library of geometric entities for `openturbine` (proposed namespace `oturb::common::geom`) and the concept of a three dimensional `Point` in the Cartesian CSYS appears to be a reasonable place to start. The specific requirements for this task are to:
- introduce a new class for `Point`s
- provide appropriate constructors to instantiate `Point` objects
- define operations on points such as distance calculation, addition/subtraction/equality of two points etc. (using operator overloading)

The development work should provide comprehensive unit tests as verification for the added logic, ideally following the test-driven development workflow.

PR #51, opened 2023-08-23
Add a data structure to represent 3-D points

faisal-bhuiyan:
Implements #16.

This work introduces a basic `Point` class to facilitate the upcoming work on geometrically exact beam theory (GEBT). Looking ahead, we will need to support construction of several geometric (e.g., vector of points to represent the beam reference line) and finite element concepts (e.g., nodes, elements etc.) and defining a `Point` class seems like a good first step. The requirements for the data structures were outlined [here](https://github.com/michaelasprague/OpenTurbineTheory/tree/main/BeamDataStructureNotes).


src/gebt_poc/point.h
```diff
@@ -0,0 +1,48 @@
+#pragma once
+
+#include "src/rigid_pendulum_poc/vector.h"
+
+using Vector = openturbine::rigid_pendulum::Vector;
```
ddement, line 5: Don't use `using` statements in the global namespace in headers.  Every file that includes this will now have this alias. Put this statement inside the point class itself to encapsulate it.

ddement (approved):
Looks good. I verified that everything passes on CUDA.

PR #53, opened 2023-08-29
WIP: Add CUDA Support to CI

ddement:


alanw0:
What's the status of this pull request? Seems like a good idea to have the CI check cuda...

ddement:
> What's the status of this pull request? Seems like a good idea to have the CI check cuda...

The problem is two-fold. For compilation, it adds a significant amount of time to the build.  We don't have many cores to work with, so building magma and Kokkos-Kernels with CUDA support is slow. 

For running, GitHub's CI runners don't have GPUs. There don't seem to be options for emulating them, so we'd need machines with the proper hardware (it's oft requested, if Google searches are to be believed).  

I think there are ways around these limitations, in theory, but they are quite complicated.  It's worth noting that neither trilinos nor kokkos test on GPU for their GitHub based CI.

alanw0:
ok, that's good information. Sounds like it's impossible then.
Does github CI have a way to let you specify a non-github machine to use for running the CI? Probably couldn't be a Sandia machine since network security issues would no doubt prevent it.
But otherwise, there may be no choice but to abandon this PR.

faisal-bhuiyan:
@ddement Should we close this PR if we're not actively working on it?

ddement:
Closing because of the discussed shortcomings of our testing infrastructure.  The lack of testing on CUDA (or other GPU platforms, for that matter) is still a problem, but one what will be addressed in the future.  Recent removal of the dependency on Magma might make a CUDA build line run in a reasonable amount of time, but this has not been tested.

PR #55, opened 2023-08-30
Add Dockerfile to automatically build a docker image of OpenTurbine

faisal-bhuiyan:
This PR adds a Dockerfile to facilitate the automatic building of a Docker image for OpenTurbine. The Docker image would allow developers/users to easily run OpenTurbine in a contained environment without the need for complex setup procedures.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,55 @@
+# Use an official Ubuntu as a parent image
+FROM ubuntu:latest
+
+# Install required dependencies
+RUN apt-get update && \
+    apt-get install -y \
+    clang-format-14 \
+    cmake \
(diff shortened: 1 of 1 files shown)


michaelasprague:
@faisal-bhuiyan -- this is not a priority.  Focus should be on getting the minimal algorithms in place for the next proof-of-concept step, which is the beam.

faisal-bhuiyan:
@michaelasprague @alanw0 Thanks for your input. I know this is not a priority, I developed it mostly for myself so I could build OpenTurbine on my mac in the linux environment. Thought it might be useful to someone else if we version controlled it in the repo.

PR #58, opened 2023-09-06
Removed dependence on relocatable device code

ddement:


alanw0 (approved):
Basically just keeping all device code in headers?
Probably reasonable, to avoid RDC.

faisal-bhuiyan (approved):
Looks good. Adding a little description on what is being done in the PR would be helpful imo.

Issue #57, opened 2023-09-06
Data structure for sectional mass and stiffness matrices


faisal-bhuiyan:
Per the requirements laid out in the [theoretical development](https://github.com/michaelasprague/OpenTurbineTheory/tree/main/BeamDataStructureNotes), we need to take in the material properties as a series of 6 x 6 sectional mass and stiffness matrices, defined in the beam materials csys. Following screenshot provides details of the requirements.

![image](https://github.com/Exawind/openturbine/assets/59463700/60d8ca96-52f2-498b-8311-6221e7c4219a)

Above can be broken down into the following development tasks:
- Mass and Stiffness matrices each are 6 x 6 and we need to establish these concepts into OpenTurbine (that does not already exist)
- Create a `Section` class that consists of the mass matrix, stiffness matrix, and a normalized location along the length of the beam
- A `std::vector<Section>` provides the complete list of sectional mass and stiffness matrices, to be used as material properties for the finite element code

PR #65, opened 2023-09-13
Introduce a data structure to handle beam sections

faisal-bhuiyan:
Takes care of the second item outlined in the tasks listed in #57.

> Create a Section class that consists of the mass matrix, stiffness matrix, and a normalized location along the length of the beam

### DIFF
--- a/src/gebt_poc/section.cpp
+++ b/src/gebt_poc/section.cpp
@@ -11,4 +11,13 @@ StiffnessMatrix::StiffnessMatrix(const Kokkos::View<double**> stiffness)
     Kokkos::deep_copy(stiffness_matrix_, stiffness);
 }
 
+Section::Section(
+    double location, gen_alpha_solver::MassMatrix mass_matrix, StiffnessMatrix stiffness_matrix
+)
+    : location_(location), mass_matrix_(mass_matrix), stiffness_matrix_(stiffness_matrix) {
+    if (location_ < 0. || location_ > 1.) {
+        throw std::invalid_argument("Section location must be between 0 and 1");
+    }
+}
+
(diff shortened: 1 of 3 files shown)


src/gebt_poc/section.cpp
```diff
@@ -11,4 +11,13 @@ StiffnessMatrix::StiffnessMatrix(const Kokkos::View<double**> stiffness)
     Kokkos::deep_copy(stiffness_matrix_, stiffness);
 }
 
+Section::Section(
+    double location, gen_alpha_solver::MassMatrix mass_matrix, StiffnessMatrix stiffness_matrix
```
alanw0, line 15: Is MassMatrix in a different namespace than StiffnessMatrix?
I don't know if it matters, but seems odd.

faisal-bhuiyan: We added the `MassMatrix` previously as part of the generalized-alpha proof-of-concept (poc) work. I'm just importing this functionality rather than duplicating code. I also find it odd, however from what I remember, looks like we should do the development work for beam poc separately?

ddement: I agree that the namespacing is weird.  I imagine that we'll have a final namespace (and some sub-namespaces?) once everything is said and done.  Let's leave it as is for now.  If it becomes too awkward in places, we can move the parts we need into the new namespace.

ddement (approved):
For functionality that you expect to be called on GPUs (i.e. inside parallel_for calls), you should decorate them with KOKKOS_FUNCTION or KOKKOS_INLINE_FUNCTION.  This will save us a few build cycles when testing it on device.  No need to change anything now, just a note for the future.

PR #67, opened 2023-09-14
Create a Model class to perform the beam FE model creation

faisal-bhuiyan:
A `Model` class was introduced in this PR with the responsibility of creating the components of a beam finite element model based on the input provided by the user. As a first step, the following attribute mentioned in #57 was added.

> A `std::vector<Section>` provides the complete list of sectional mass and stiffness matrices, to be used as material properties for the finite element code

### DIFF
--- a/src/gebt_poc/model.cpp
+++ b/src/gebt_poc/model.cpp
@@ -2,24 +2,11 @@
 
 namespace openturbine::gebt_poc {
 
-Kokkos::View<Section*> CreateSections(std::vector<Section> sections) {
-    auto sections_view = Kokkos::View<Section*>("sections", sections.size());
-    auto sections_host = Kokkos::create_mirror(sections_view);
-
-    for (std::size_t index = 0; index < sections.size(); ++index) {
-        sections_host(index) = sections[index];
-    }
-
-    // *** Question for David: How to implement a deep_copy() for a Kokkos::View<Section*>?
-    Kokkos::parallel_for(
-        sections.size(),
-        KOKKOS_LAMBDA(std::size_t index) { sections_view(index) = sections_host(index); }
-    );
-
-    return sections_view;
+Model::Model(std::string name) : name_(name) {
 }
 
-Model::Model(std::string name, Kokkos::View<Section*> sections) : name_(name), sections_(sections) {
+Model::Model(std::string name, std::vector<Section> sections)
+    : name_(name), sections_(std::move(sections)) {
 }
 
 }  // namespace openturbine::gebt_poc
--- a/src/gebt_poc/model.cpp
+++ b/src/gebt_poc/model.cpp
@@ -2,9 +2,24 @@
 
 namespace openturbine::gebt_poc {
 
-Model::Model(std::string name, Kokkos::View<Section*> sections)
-    : name_(name), sections_("sections", sections.extent(0)) {
-    Kokkos::deep_copy(sections_, sections);
+Kokkos::View<Section*> CreateSections(std::vector<Section> sections) {
+    auto sections_view = Kokkos::View<Section*>("sections", sections.size());
+    auto sections_host = Kokkos::create_mirror(sections_view);
+
+    for (std::size_t index = 0; index < sections.size(); ++index) {
+        sections_host(index) = sections[index];
+    }
+
+    // *** Question for David: How to implement a deep_copy() for a Kokkos::View<Section*>?
+    Kokkos::parallel_for(
+        sections.size(),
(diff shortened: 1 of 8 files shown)


src/gebt_poc/model.h
```diff
@@ -0,0 +1,26 @@
...
+
+    /// Adds a section to the model
+    void AddSection(Section section) { sections_.emplace_back(section); }
+
+    /// Returns the sections of the model
+    inline const std::vector<Section>& GetSections() const { return sections_; }
+
+private:
+    std::string name_;               //< Name of the model
+    std::vector<Section> sections_;  //< Sections of the beam
```
ddement, line 23: Should our Sections be stored in a vector?  std::vector can't be used on GPUs, so we'll need to somehow make this data accessible there.

faisal-bhuiyan: I converted the above into a `Kokkos::View<Section*>`, along with other necessary changes required by it.

src/gebt_poc/model.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+Kokkos::View<Section*> CreateSections(std::vector<Section> sections) {
+    auto sections_view = Kokkos::View<Section*>("sections", sections.size());
+    auto sections_host = Kokkos::create_mirror(sections_view);
+
+    for (std::size_t index = 0; index < sections.size(); ++index) {
+        sections_host(index) = sections[index];
+    }
+
+    // *** Question for David: How to implement a deep_copy() for a Kokkos::View<Section*>?
```
faisal-bhuiyan, line 13: @ddement Please take a look at this the next time you review it. Looks like we need to implement a `deep_copy()` function for the `Section` class to be able to simply write `deep_copy(sections_view, sections_host)`. Do you think it's something we should look into implementing?

src/gebt_poc/section.h
```diff
@@ -20,19 +23,24 @@ class StiffnessMatrix {
...
 
     /// Returns the mass matrix of the section
-    inline const gen_alpha_solver::MassMatrix& GetMass() const { return mass_matrix_; }
+    inline const gen_alpha_solver::MassMatrix& GetMassMatrix() const { return mass_matrix_; }
 
     /// Returns the stiffness matrix of the section
     inline const StiffnessMatrix& GetStiffnessMatrix() const { return stiffness_matrix_; }
 
 private:
+    std::string name_;                          //< Name of the section
```
ddement, line 43: Why do sections need a name?

src/gebt_poc/model.cpp
ddement: On further reflection, it might be smart to not have Views of sections.  Perhaps it would make more sense to just store our matrices as `View<double***>` instances with the first dimension being the number of sections and the second/third dimensions being the matrix sizes.  If it makes sense, we could do a "Sections" class (probably a bad name) to hold the mass, stiffness, and location views.  I think doing a "structure of arrays" rather than "array of structures" in this case will work much better with the GPU.  If you ever want to pass a single matrix into a function, you can use a `Kokkos::subview`.  Does this make sense?  Thoughts?

What you have implemented here wouldn't work on GPU (sections_host can't be accessed on GPU).  We'd either want to implement the deep_copy as you said (I'd have to figure out how to do that), or copy parts of the section into stack memory so that it could be captured up to device.

src/gebt_poc/section.h
faisal-bhuiyan: It's a typical feature in FEA codes, imo. We might have different sections that are assigned to different elements of the model and it's a quick check to assert we have the right material/section assigned by checking the name.

src/gebt_poc/model.cpp
faisal-bhuiyan: @ddement I finally had a chance to think about the data structure to be used for model creation based on user inputs. I see your point regarding the SoA vs AoS and I agree it will be more efficient for working with GPUs. However, let's remember the `Section`s being created here are just a placeholder for user provided inputs (at some random points) and will be used later on to create the sectional properties at the quadrature points via the interpolation scheme - those are the data structures that will be used in the time integration scheme and we should look into making compatible with GPU. 

For the purposes of this PR, I think we could just create a `std::vector<Section>`, much like some of the recent work where we are leveraging the `std::` containers and algorithms. These are the pre-processing steps and we could run them only in CPU. Does that sound reasonable to you?

@michaelasprague Please take a look and free to let me know if I'm missing something in the above.

faisal-bhuiyan:
@ddement Let's review this one more time to make sure everything looks good please.

ddement (approved):

PR #68, opened 2023-09-16
Implement nearest neighbor search and interpolate two matrices linearly

### LINKED ISSUE #62: Enable linear interpolation of sectional mass and stiffness matrices
We need to implement a matrix interpolation infrastructure to support the interpolation of mass and stiffness matrices from user-provided locations to quadrature points of the element or other points as necessary. Linear interpolation of nearest neighbors seems like a good first strategy for the interpolation scheme. The following excerpt from the requirements document capture the details of the work.

image

faisal-bhuiyan:
Fixes #62.

Two functionalities were added in this work
- Search for k-nearest neighbors of a point from a given list of points
- Linearly interpolate two matrices

### DIFF
--- a/src/gebt_poc/interpolation.cpp
+++ b/src/gebt_poc/interpolation.cpp
@@ -1,0 +1,19 @@
+#include "src/gebt_poc/interpolation.h"
+
+namespace openturbine::gebt_poc {
+
+Point FindNearestNeighbor(const std::vector<Point>& points, const Point& pt) {
+    if (points.empty()) {
+        throw std::invalid_argument("Points list must not be empty");
+    }
+
+    // Find the nearest neighbor using std algorithm
+    auto nearest_neighbor =
+        std::min_element(points.begin(), points.end(), [&pt](const Point& lhs, const Point& rhs) {
+            return pt.DistanceTo(lhs) < pt.DistanceTo(rhs);
+        });
+
+    return *nearest_neighbor;
+}
+
+}  // namespace openturbine::gebt_poc
--- a/src/gebt_poc/interpolation.cpp
+++ b/src/gebt_poc/interpolation.cpp
@@ -39,4 +39,21 @@ std::vector<Point> FindkNearestNeighbors(
     return k_nearest_neighbors;
 }
 
+Kokkos::View<double**> LinearInterpolation(
+    const Kokkos::View<double**> m1, const Kokkos::View<double**> m2, const double alpha
+) {
+    if (m1.extent(0) != m2.extent(0) || m1.extent(1) != m2.extent(1)) {
+        throw std::invalid_argument("Matrices must have the same dimensions");
+    }
+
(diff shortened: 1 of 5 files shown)


alanw0 (approved):
If we only want this search to operate on the host, then it looks good to me.

src/gebt_poc/interpolation.cpp
```diff
@@ -0,0 +1,59 @@
...
+    if (m1.extent(0) != m2.extent(0) || m1.extent(1) != m2.extent(1)) {
+        throw std::invalid_argument("Matrices must have the same dimensions");
+    }
+
+    Kokkos::View<double**> interpolated_matrix("interpolated_matrix", m1.extent(0), m1.extent(1));
+    for (size_t i = 0; i < m1.extent(0); i++) {
+        for (size_t j = 0; j < m1.extent(1); j++) {
+            interpolated_matrix(i, j) = (1. - alpha) * m1(i, j) + alpha * m2(i, j);
+        }
+    }
```
ddement, line 54: These views all live on GPU.  They must be accessed from inside a parallel_for call, otherwise we'll get SegFaults when building for GPU.  I'd propose just enclosing these lines in `Kokkos::parallel_for(1, KOKKOS_LAMBDA(int){...})` since the dimensions of m are so small that parallelization doesn't make much sense.

tests/unit_tests/gebt_poc/test_interpolation.cpp
```diff
@@ -0,0 +1,198 @@
...
+    auto nearest_neighbors_4 = FindkNearestNeighbors(points, pt_4, 2);
+    std::vector<Point> expected_4 = {Point(9., 0., 0.), Point(8., 0., 0.)};
+    EXPECT_EQ(nearest_neighbors_4, expected_4);
+}
+
+TEST(MatrixInterpolationTest, LinearlyInterpolate1x1Matrices) {
+    Kokkos::View<double**> m1("m1", 1, 1);
+    Kokkos::View<double**> m2("m2", 1, 1);
+    m1(0, 0) = 1.;
+    m2(0, 0) = 2.;
```
ddement, line 103: When setting the values of a View, either create a host mirror and copy or do so inside a prallel_for call.

src/gebt_poc/interpolation.cpp
```diff
@@ -0,0 +1,59 @@
...
+                remaining_pts.begin(), remaining_pts.end(),
+                [&nn](const Point& p) {
+                    return p == nn;
+                }
+            ),
+            remaining_pts.end()
+        );
+    }
+    return k_nearest_neighbors;
+}
```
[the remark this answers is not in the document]
faisal-bhuiyan: I spent some time looking into this, however it looks like currently Kokkos implementation of the std algorithms only accept 1D Views. Since we are sorting a list of `Point`s, a 2D view would be required. Feel free to let me know if I'm missing something here.

image

PR #69, opened 2023-09-22
Generate Gauss-Lobatto-Legendre points

### LINKED ISSUE #61: Introduce the concept of polynomial shape functions
To support the Legendre Spectral Finite Element (LSFE) analysis of a beam, we need to implement the concept of Lagrange polynomial shape functions (aka interpolation functions). Details of the requirement are provided in the following.

image

faisal-bhuiyan:
Partially resolves #61.

In this work, Gauss-Lobatto-Legendre (GLL) points required for nodal locations using polynomial shape/interpolation functions of order n.

### DIFF
--- a/src/gebt_poc/interpolation.cpp
+++ b/src/gebt_poc/interpolation.cpp
@@ -1,10 +1,12 @@
 #include "src/gebt_poc/interpolation.h"
 
+#include "src/utilities/log.h"
+
 namespace openturbine::gebt_poc {
 
 Point FindNearestNeighbor(const std::vector<Point>& points_list, const Point& point) {
     if (points_list.empty()) {
-        throw std::invalid_argument("points_list list must not be empty");
+        throw std::invalid_argument("points list must not be empty");
     }
 
     auto nearest_neighbor = std::min_element(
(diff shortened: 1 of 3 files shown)


ddement (approved):
Looks good. 

One question: our GLL points are all colinear.  Is there a reason why we're generating points from this function?  Couldn't we just create a vector of doubles in this case?  Of course, it will be transformed into 3D when we use it in physical space, but this function is just generating stuff in reference space.

michaelasprague:
@ddement  -- These beam elements are fully 3D.  However, all elements are mapped onto the reference line with coordinate -1 \le \xi \le 1; so we only need the 1D GLL points.  Does that address your question?

faisal-bhuiyan:
> Looks good.
> 
> One question: our GLL points are all colinear. Is there a reason why we're generating points from this function? Couldn't we just create a vector of doubles in this case? Of course, it will be transformed into 3D when we use it in physical space, but this function is just generating stuff in reference space.

As you pointed out, we could be generating just doubles as output from `GenerateGLLPoints` function. I opted to output `Point`s to be able to use any logic from that class and not duplicate code. Does that make sense?

ddement:
> @ddement -- These beam elements are fully 3D. However, all elements are mapped onto the reference line with coordinate -1 \le \xi \le 1; so we only need the 1D GLL points. Does that address your question?

My point was, since this function is only generating the GLL points on the reference line, it could be represented as a `std::vector<double>` rather than a `std::vector<Point>` where two components are set to 0 by definition.  The current code should work fine, I just wanted to make sure that I wasn't missing anything.

faisal-bhuiyan:
> > @ddement -- These beam elements are fully 3D. However, all elements are mapped onto the reference line with coordinate -1 \le \xi \le 1; so we only need the 1D GLL points. Does that address your question?
> 
> My point was, since this function is only generating the GLL points on the reference line, it could be represented as a `std::vector<double>` rather than a `std::vector<Point>` where two components are set to 0 by definition. The current code should work fine, I just wanted to make sure that I wasn't missing anything.

I opted for keeping things simpler and use a `std::vector<double>` to represent the GLL points. It's submitted as part of the [shape function PR](https://github.com/Exawind/openturbine/pull/70), let's go ahead and merge this one (once the base branches for this are merged in).

PR #76, opened 2023-09-28
Adds a simple mesh data structure for beam elements and a factory function

ddement:
This data structure provides an association of elements to nodes, so that we can loop elements and extract which nodes are owned by each one.  State data (positions, velocities, etc) will be stored externally and reference this structure for sizing/access.  The key assembly loops will be parallel over elements, then gather/scatter based on nodal indices.

This is a very simple mesh, but it should meet our needs for now.  One requirement is that all elements have the same number of nodes.

### DIFF
--- a/src/gebt_poc/mesh.h
+++ b/src/gebt_poc/mesh.h
@@ -3,49 +3,45 @@
 #include <Kokkos_Core.hpp>
 
 namespace openturbine {
-  class Mesh {
-  public:
-    int GetNumberOfElements() {
-      return number_of_elements_;
-    }
+class Mesh {
+public:
+    int GetNumberOfElements() { return number_of_elements_; }
 
-    int GetNumberOfNodes() {
-      return number_of_nodes_;
-    }
+    int GetNumberOfNodes() { return number_of_nodes_; }
 
     void SetNumberOfElements(int number_of_elements) { number_of_elements_ = number_of_elements; }
     void SetNumberOfNodes(int number_of_nodes) { number_of_nodes_ = number_of_nodes; }
 
     void InitializeElementNodeConnectivity(int number_of_elements, int nodes_per_element) {
-      connectivity_ = Kokkos::View<int**>("connectivity", number_of_elements, nodes_per_element);
+        connectivity_ = Kokkos::View<int**>("connectivity", number_of_elements, nodes_per_element);
     }
 
     KOKKOS_FUNCTION
     Kokkos::View<const int*> GetNodesForElement(int elementID) {
-      return Kokkos::subview(connectivity_, elementID, Kokkos::ALL);
+        return Kokkos::subview(connectivity_, elementID, Kokkos::ALL);
     }
 
     friend Mesh create1DMesh(int number_of_elements, int nodes_per_element);
-  protected:
+
+protected:
     int number_of_elements_;
     int number_of_nodes_;
     Kokkos::View<int**> connectivity_;
-  };
+};
 
-  
-  inline Mesh create1DMesh(int number_of_elements, int nodes_per_element) {
+inline Mesh create1DMesh(int number_of_elements, int nodes_per_element) {
     Mesh mesh;
     mesh.SetNumberOfElements(number_of_elements);
     mesh.SetNumberOfNodes(number_of_elements * nodes_per_element - (number_of_elements - 1));
     mesh.InitializeElementNodeConnectivity(number_of_elements, nodes_per_element);
 
     auto set_element_connectivity = KOKKOS_LAMBDA(int elementID) {
-      for(int node = 0; node < nodes_per_element; ++node) {
-        mesh.connectivity_(elementID, node) = node + (nodes_per_element-1)*(elementID);
-      }
+        for (int node = 0; node < nodes_per_element; ++node) {
+            mesh.connectivity_(elementID, node) = node + (nodes_per_element - 1) * (elementID);
+        }
     };
 
     Kokkos::parallel_for(number_of_elements, set_element_connectivity);
     return mesh;
-  }
 }
+}  // namespace openturbine
--- a/src/gebt_poc/mesh.h
+++ b/src/gebt_poc/mesh.h
@@ -1,0 +1,51 @@
+#pragma once
+
+#include <Kokkos_Core.hpp>
+
+namespace openturbine {
+  class Mesh {
+  public:
+    int GetNumberOfElements() {
+      return number_of_elements_;
+    }
+
+    int GetNumberOfNodes() {
+      return number_of_nodes_;
+    }
+
+    void SetNumberOfElements(int number_of_elements) { number_of_elements_ = number_of_elements; }
+    void SetNumberOfNodes(int number_of_nodes) { number_of_nodes_ = number_of_nodes; }
+
+    void InitializeElementNodeConnectivity(int number_of_elements, int nodes_per_element) {
+      connectivity_ = Kokkos::View<int**>("connectivity", number_of_elements, nodes_per_element);
+    }
+
+    KOKKOS_FUNCTION
+    Kokkos::View<const int*> GetNodesForElement(int elementID) {
+      return Kokkos::subview(connectivity_, elementID, Kokkos::ALL);
+    }
+
+    friend Mesh create1DMesh(int number_of_elements, int nodes_per_element);
+  protected:
+    int number_of_elements_;
+    int number_of_nodes_;
+    Kokkos::View<int**> connectivity_;
+  };
+
+  
+  inline Mesh create1DMesh(int number_of_elements, int nodes_per_element) {
+    Mesh mesh;
+    mesh.SetNumberOfElements(number_of_elements);
+    mesh.SetNumberOfNodes(number_of_elements * nodes_per_element - (number_of_elements - 1));
+    mesh.InitializeElementNodeConnectivity(number_of_elements, nodes_per_element);
+
+    auto set_element_connectivity = KOKKOS_LAMBDA(int elementID) {
+      for(int node = 0; node < nodes_per_element; ++node) {
+        mesh.connectivity_(elementID, node) = node + (nodes_per_element-1)*(elementID);
+      }
+    };
+
+    Kokkos::parallel_for(number_of_elements, set_element_connectivity);
+    return mesh;
+  }
+}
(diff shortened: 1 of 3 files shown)


src/gebt_poc/mesh.h
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    int GetNumberOfNodes() {
+      return number_of_nodes_;
+    }
+
+    void SetNumberOfElements(int number_of_elements) { number_of_elements_ = number_of_elements; }
+    void SetNumberOfNodes(int number_of_nodes) { number_of_nodes_ = number_of_nodes; }
+
+    void InitializeElementNodeConnectivity(int number_of_elements, int nodes_per_element) {
```
alanw0, line 15: Maybe there should be a debug-mode assert here to make sure the number of elements and total nodes will end up consistent with the separately-settable values that the caller can give to `SetNumberOfElements` and `SetNumberOfNodes`.

alanw0 (approved):
Will there be 2D and 3D meshes? Or only 1D?
In any case, letting the mesh be simply the connectivity table seems like a good start.

ddement:
> Will there be 2D and 3D meshes? Or only 1D? In any case, letting the mesh be simply the connectivity table seems like a good start.

We've only discussed beam elements so far.  This should support beams in 2D/3D just fine as well as arbitrary structures made from those beams.  We'll need other factory functions to build them, as this factory is just a number of elements lined up end to end.  

If we need to add more complexity (quads, hexes, etc), we'll want a more complete solution, be that STK or at least incorporating some of those formalisms.

src/gebt_poc/mesh.h
ddement: Checking element consistency should be trivial. Checking the consistency of the nodes will involve some reductions, since some nodes can exist in multiple elements.  Just checking if the maximum node index referenced in the connectivity is less than the number of nodes should be easy enough.

src/gebt_poc/mesh.h
```diff
@@ -0,0 +1,47 @@
...
+
+    auto set_element_connectivity = KOKKOS_LAMBDA(int elementID) {
+        for (int node = 0; node < nodes_per_element; ++node) {
+            mesh.connectivity_(elementID, node) = node + (nodes_per_element - 1) * (elementID);
+        }
+    };
+
+    Kokkos::parallel_for(number_of_elements, set_element_connectivity);
+    return mesh;
+}
```
faisal-bhuiyan, line 46: I might be missing something here, but I don't see why we need a separate class for creating 1D mesh. Couldn't we go through the interface of `Mesh` class and instantiate a 1D/2D/3D mesh as necessary?

src/gebt_poc/mesh.h
```diff
@@ -0,0 +1,47 @@
+#pragma once
+
+#include <Kokkos_Core.hpp>
+
+namespace openturbine {
```
faisal-bhuiyan, line 5: The namespace should be `namespace openturbine::gebt_poc` to be consistent.

faisal-bhuiyan (commented):
Overall, the data structure of `Mesh` looks good to me. I have a suggestion regarding the design of `Mesh` class and a friend `1DMesh` class to instantiate it. Please see the line-by-line comments.

src/gebt_poc/mesh.h
ddement: Since the user has to go through these functions, we can ensure that a valid mesh is created from a simple UI where opportunities to make an invalid mesh are encapsulated.  As we need different mesh topologies, we can add simple functions to make them or we can add functions that read them from other input sources. 

There are plenty of features that we could add to make things more flexible or different ways to do it, but I find this one both an simple solution to the problem at hand and easily extended to the problems in our immediate future.

src/gebt_poc/mesh.h
```diff
@@ -0,0 +1,47 @@
...
+    Kokkos::View<const int*> GetNodesForElement(int elementID) {
+        return Kokkos::subview(connectivity_, elementID, Kokkos::ALL);
+    }
+
+    friend Mesh create1DMesh(int number_of_elements, int nodes_per_element);
+
+protected:
+    int number_of_elements_;
+    int number_of_nodes_;
+    Kokkos::View<int**> connectivity_;
```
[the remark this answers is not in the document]
ddement: I doubt that we'll ever need the extra numbers that we get from using an unsigned type, and signed integers have better usability (for example, overflow/underflow can be checked quickly, backwards iteration is less error prone).  I tend to prefer int except when an unsigned type is required for correctness or avoiding warnings.

src/gebt_poc/mesh.h
```diff
@@ -0,0 +1,47 @@
...
+class Mesh {
+public:
+    int GetNumberOfElements() { return number_of_elements_; }
+
+    int GetNumberOfNodes() { return number_of_nodes_; }
+
+    void SetNumberOfElements(int number_of_elements) { number_of_elements_ = number_of_elements; }
+    void SetNumberOfNodes(int number_of_nodes) { number_of_nodes_ = number_of_nodes; }
+
+    void InitializeElementNodeConnectivity(int number_of_elements, int nodes_per_element) {
```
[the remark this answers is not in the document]
ddement: I've moved these functions into the protected namespace and actually plan to do so with the default constructor (I thought I had already done that).  I want to avoid depending on the details of how I've implemented this mesh except in very intentional locations.  It handles the simple case of beam elements with a fixed number of nodes very well, but will need to be changed quite a bit to do other things.  It's better to control the leakage of internal structure now rather than have to clean things up in the future.

alanw0 (approved):
Yes I like the factory function, that's a good way to control/validate inputs, and it's a good way to add extra features or mesh configurations later with more factory functions.

src/gebt_poc/mesh.h
ddement: I added a function to do some basic checks.  There probably are more checks to do to ensure consistency, so we can add more as we run into those edge cases.  Mesh construction will be happening under controlled conditions and the meshes will be pretty simple for this milestone, so I don't think we need much more at the moment.  By the time we get to more complex cases, I am assuming we'll need a more sophisticated data structure to handle them.

src/gebt_poc/mesh.h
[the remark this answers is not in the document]
faisal-bhuiyan: ok fair enough, not a huge deal either way

src/gebt_poc/mesh.h
```diff
@@ -0,0 +1,100 @@
...
+        CheckNodeIDConsistency();
+        CheckNodeUniquenessConsistency();
+    }
+
+    int number_of_elements_;
+    int number_of_nodes_;
+    Kokkos::View<int**> connectivity_;
+};
+
+inline Mesh create1DMesh(int number_of_elements, int nodes_per_element) {
```
faisal-bhuiyan, line 83: Let's rename this function -> `Create1DMesh` to be consistent with the style.

src/gebt_poc/mesh.h
```diff
@@ -0,0 +1,100 @@
...
+        CheckElementConsistency();
+        CheckNodeIDConsistency();
+        CheckNodeUniquenessConsistency();
+    }
+
+    int number_of_elements_;
+    int number_of_nodes_;
+    Kokkos::View<int**> connectivity_;
+};
+
```
faisal-bhuiyan, line 82: Let's consider adding a doxygen comment for the intent of this function, e.g.

`/// Creates a 1D mesh with element connectivity in increasing order from left to right`

src/gebt_poc/mesh.h
```diff
@@ -0,0 +1,100 @@
...
+    void CheckConsistency() {
+        CheckElementConsistency();
+        CheckNodeIDConsistency();
+        CheckNodeUniquenessConsistency();
+    }
+
+    int number_of_elements_;
+    int number_of_nodes_;
+    Kokkos::View<int**> connectivity_;
```
faisal-bhuiyan, line 80: Let's add a description for this data attribute which quickly communicates it's intent.

src/gebt_poc/mesh.h
```diff
@@ -0,0 +1,100 @@
...
+                node_list_vector[node] = host_node_list(node);
+            }
+            std::sort(std::begin(node_list_vector), std::end(node_list_vector));
+            auto first_duplicate =
+                std::adjacent_find(std::begin(node_list_vector), std::end(node_list_vector));
+            if (first_duplicate != end(node_list_vector)) {
+                throw std::domain_error("Nodes within an element must be unique");
+            }
+        }
+    }
```
faisal-bhuiyan, line 70: Let's add one line description for these checks that communicates their intent to the reader.

src/gebt_poc/mesh.h
```diff
@@ -0,0 +1,100 @@
...
+
+namespace openturbine::gebt_poc {
+class Mesh {
+public:
+    int GetNumberOfElements() const { return number_of_elements_; }
+
+    int GetNumberOfNodes() const { return number_of_nodes_; }
+
+    KOKKOS_FUNCTION
+    Kokkos::View<const int*> GetNodesForElement(int elementID) const {
```
faisal-bhuiyan, line 13: Let's rename `elementID` -> `element_id` or `elem_id` to be consistent with the style guide.

src/gebt_poc/mesh.h
```diff
@@ -0,0 +1,100 @@
...
+    Mesh() = default;
+    void SetNumberOfElements(int number_of_elements) { number_of_elements_ = number_of_elements; }
+    void SetNumberOfNodes(int number_of_nodes) { number_of_nodes_ = number_of_nodes; }
+
+    void InitializeElementNodeConnectivity(int number_of_elements, int nodes_per_element) {
+        connectivity_ = Kokkos::View<int**>("connectivity", number_of_elements, nodes_per_element);
+    }
+
+    void CheckElementConsistency() {
+        if (static_cast<unsigned>(number_of_elements_) != connectivity_.extent(0)) {
```
faisal-bhuiyan, line 29: Let's do the type casting to `std::size_t` to be consistent with rest of the code in this file.
`if (static_cast<std::size_t>(number_of_elements_) != connectivity_.extent(0)) {`

faisal-bhuiyan (approved):
LGTM. Only several minor comments below.

src/gebt_poc/mesh.h
ddement: The error messages already provide a description of the error checked.

PR #77, opened 2023-09-29
Calculate the jacobian for an element

faisal-bhuiyan:
Provides an implementation for #75.

### DIFF
--- a/src/gebt_poc/element.cpp
+++ b/src/gebt_poc/element.cpp
@@ -15,10 +15,6 @@ double CalculateJacobian(
         return result;
     };
 
-    auto two_norm = [](const std::array<double, 3>& vector) {
-        return std::sqrt(std::pow(vector[0], 2) + std::pow(vector[1], 2) + std::pow(vector[2], 2));
-    };
-
     // Get all the x, y, z components of the nodes in three separate vectors
     std::vector<double> x_components{};
     std::vector<double> y_components{};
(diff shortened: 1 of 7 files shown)


ddement (approved):

src/gebt_poc/element.cpp
```diff
@@ -0,0 +1,43 @@
...
+) {
+    auto dot_product = [](const std::vector<double>& vector_1, const std::vector<double>& vector_2) {
+        auto result = 0.;
+        for (size_t i = 0; i < vector_1.size(); ++i) {
+            result += vector_1[i] * vector_2[i];
+        }
+        return result;
+    };
+
+    auto two_norm = [](const std::array<double, 3>& vector) {
```
alanw0, line 18: it's a bit confusing to use the argument name 'vector' when the type is array.

src/gebt_poc/element.cpp
```diff
@@ -0,0 +1,43 @@
...
+        return std::sqrt(std::pow(vector[0], 2) + std::pow(vector[1], 2) + std::pow(vector[2], 2));
```
alanw0, line 19: didn't you write a math vector recently? why not use that here for these dot and norm calculations?

alanw0 (approved):

src/gebt_poc/element.cpp
faisal-bhuiyan: Thanks for catching this - the `Vector` class is a data structure for 3 component physical vectors only, so we could use this for the norm calculation of the Jacobian, but not the dot product calculation since the `std::vector`s consist of varying # of components. Let me know if above doesn't make sense to you.

PR #83, opened 2023-10-18
Implement static iteration matrix for a single beam element

### LINKED ISSUE #73: Implement tangent stiffness for iteration matrix for single static beam

faisal-bhuiyan:
Resolves #73.

Namely the iteration matrix (under static loading) for a beam element was implemented based on the development provided in the accompanying [theory guide](https://github.com/michaelasprague/OpenTurbineTheory/tree/main/BeamTheoryNotes) and unit tests provided in the [mathematica implementation](https://github.com/michaelasprague/OpenTurbineTheory/blob/main/mathematica/StaticIterationMatrix.nb). 

As a follow up work, we need to adopt this to work with our field and mesh data structures.


src/gebt_poc/solver.cpp
```diff
@@ -234,4 +234,303 @@ Kokkos::View<double*> CalculateStaticResidual(
...
+            {0, 0}, {kNumberOfVectorComponents, kNumberOfVectorComponents}
+        ),
+        KOKKOS_LAMBDA(const size_t i, const size_t j) {
+            O_matrix(i, j) = 0.;
+            O_matrix(i + kNumberOfVectorComponents, j) = 0.;
+            O_matrix(i, j + kNumberOfVectorComponents) = non_zero_terms_part_1(i, j);
+            O_matrix(i + kNumberOfVectorComponents, j + kNumberOfVectorComponents) =
+                non_zero_terms_part_2(i, j);
+        }
+    );
```
ddement, line 323: Perhaps each of the O, P, and Q matrices should be assembled in their own function?  This would help with independent testing and make things a little more readable than a single big function.

tests/unit_tests/gebt_poc/test_solver.cpp
```diff
@@ -390,4 +390,490 @@ TEST(SolverTest, CalculateStaticResidualNonZeroValues) {
     );
 }
 
+TEST(SolverTest, CalculateIterationMatrixComponents) {
```
ddement, line 393: I think we might want some "null" tests like we have for the residual matrix.  This will make refactoring a bit easier.

ddement (approved):
Looks good.  Is there some way to move some of our expected results into a header file?  These matrices are getting pretty big to the point that they make the tests less readable.  Even something as simple as making them global variables under a "golds" namespace or something might be enough for now.

tests/unit_tests/gebt_poc/test_solver.cpp
```diff
@@ -508,7 +475,50 @@ TEST(SolverTest, CalculateIterationMatrixComponents) {
...
+TEST(SolverTest, CalculateStaticIterationMatrixWithZeroValues) {
+    // 5 nodes on a line
+    auto position_vectors = gen_alpha_solver::create_vector({
+        0., 0., 0., 1., 0., 0., 0.,  // node 1
+        1., 0., 0., 1., 0., 0., 0.,  // node 2
+        2., 0., 0., 1., 0., 0., 0.,  // node 3
+        3., 0., 0., 1., 0., 0., 0.,  // node 4
+        4., 0., 0., 1., 0., 0., 0.   // node 5
+    });
+
```
alanw0, line 487: I'm having trouble knowing when there should be 6 values per row (per node) vs when there should be 7 values per row. But that's probably not a significant issue, probably just me not keeping up with the math development lately.

faisal-bhuiyan: @alanw0 Yup, this is a direct consequence of the SO3-based mathematical development we are using. Please take a look at the following excerpt - basically position and displacements would contain 7 components/node, while velocity, acceleration will contain 6 components/node.

image

faisal-bhuiyan: Please feel free to ask more question - we should get one approval before we merge this PR

alanw0 (approved):

PR #85, opened 2023-11-02
Calculate the constraints residual vector and the constraints gradient matrix

### LINKED ISSUE #84: Implement Constraint Gradient for iteration matrix for single static beam

faisal-bhuiyan:
Fixes #84.


deslaughter:
There are a lot of Kokkos views being created in the `ConstraintsResidualVector`. Each call to `create_vector` allocates a new view and performs a deep copy of the data into the view. This may be trivial when the memory space is on the CPU, but when the memory is on the GPU, I expect there will be a large performance penalty. Could these views be contained in a structure so they could be allocated once? Does their size change once the model is set up? Kokkos is using reference counting to track the memory for these views, which are created for each constraint in each convergence iteration of the generalized alpha solve. Is there a way to make this more efficient?

Why is a view being created to pass the quaternion components to create a rotation matrix? This seems like unnecessary overhead when `EulerParameterToRotationMatrix` could take 4 arguments.

Also, using `parallel_for` for 3 elements seems inefficient. I expect there will overhead to running parallel operations which won't be negligible unless the number of elements is much larger.

Why is the data type for a rotation matrix `Kokkos::View<double **>` instead of `Kokkos::View<double[3][3]>`?

Why create new views and copy `gen_cords(0:3)` and `position_vector` instead of using them directly when calculating `x0_u0`? If it's being done for clarity, the Kokkos documentation says that the views are effectively pointers so you could do the following and avoid allocating new views.

`auto x0 = position_vector;`
`auto u0 = Kokkos::subview(gen_coords, Kokkos::make_pair(0,3));`

src/gebt_poc/solver.cpp
```diff
@@ -557,4 +557,67 @@ Kokkos::View<double**> CalculateStaticIterationMatrix(
     return iteration_matrix;
 }
 
+Kokkos::View<double*> ConstraintsResidualVector(
+    const Kokkos::View<double*> gen_coords, const Kokkos::View<double*> position_vector
+) {
+    auto rotation_0 = gen_alpha_solver::EulerParameterToRotationMatrix(
+        gen_alpha_solver::create_vector({gen_coords(3), gen_coords(4), gen_coords(5), gen_coords(6)})
```
deslaughter, line 564: Avoid allocating a new view by creating a subview of `gen_coords` containing the quaternion and passing it to `EulerParameterToRotationMatrix`:

```c++
    // Create rotation matrix from gen coordinates quaternion
    auto q0 = Kokkos::subview(gen_coords, Kokkos::make_pair(3, 7));
    auto rotation_0 = gen_alpha_solver::EulerParameterToRotationMatrix(q0);
```

src/gebt_poc/solver.cpp
```diff
@@ -557,4 +557,67 @@ Kokkos::View<double**> CalculateStaticIterationMatrix(
...
+    );
+    auto x0 =
```
deslaughter, line 566: `x0` can be a subview of `position_vector` instead of a copy:

```c++
    // Extract x0 from position vector
    auto x0 = Kokkos::subview(position_vector, Kokkos::make_pair(0, 3));
```

src/gebt_poc/solver.cpp
```diff
@@ -557,4 +557,67 @@ Kokkos::View<double**> CalculateStaticIterationMatrix(
...
+        gen_alpha_solver::create_vector({position_vector(0), position_vector(1), position_vector(2)}
+        );
+    auto u0 = gen_alpha_solver::create_vector({gen_coords(0), gen_coords(1), gen_coords(2)});
```
deslaughter, line 569: Copy `u0` from `gen_coords` directly into `constraint_residual` since `u0` is a subview:

```c++
    // Copy u0 into constraint residual from gen coordinates
    Kokkos::deep_copy(u0, Kokkos::subview(gen_coords, Kokkos::make_pair(0, 3)));
```

src/gebt_poc/solver.cpp
```diff
@@ -557,4 +557,67 @@ Kokkos::View<double**> CalculateStaticIterationMatrix(
...
+    auto x0_u0 = Kokkos::View<double*>("x0_u0", kNumberOfVectorComponents);
```
deslaughter, line 570: Use `KokkosBlas` to add `x0` and `u0`:

```c++
    // Create storage for x0 + u0, copy x0 into x0_u0, add u0 to x0
    auto x0_u0 = Kokkos::View<double*>("x0_u0", kNumberOfVectorComponents);
    Kokkos::deep_copy(x0_u0, x0);
    KokkosBlas::axpy(1.0, u0, x0_u0);
```

src/gebt_poc/solver.cpp
```diff
@@ -557,4 +557,67 @@ Kokkos::View<double**> CalculateStaticIterationMatrix(
...
+    Kokkos::parallel_for(
+        kNumberOfVectorComponents, KOKKOS_LAMBDA(const size_t i) { x0_u0(i) = x0(i) + u0(i); }
+    );
+    auto R_x0u0 = gen_alpha_solver::multiply_matrix_with_vector(rotation_0, x0_u0);
```
deslaughter, line 574: Use `KokkosBlas` to multiply the rotation matrix by `x0_u0` and subtract `x0` using `gemv`. This avoids several allocations and two `parallel_for` calls. It also avoids the reimplementation of matrix-vector multiply. The result is stored directly in the `constraint_residual` since `R_x0u0_x0` is a subview.

```c++
    // Copy x0 into R_x0u0_x0 and calculate R*x0u0-x0 with gemv
    Kokkos::deep_copy(R_x0u0_x0, x0);
    KokkosBlas::gemv("N", 1.0, rotation_0, x0_u0, -1.0, R_x0u0_x0);
```

faisal-bhuiyan:
> There are a lot of Kokkos views being created in the `ConstraintsResidualVector`. Each call to `create_vector` allocates a new view and performs a deep copy of the data into the view. This may be trivial when the memory space is on the CPU, but when the memory is on the GPU, I expect there will be a large performance penalty. Could these views be contained in a structure so they could be allocated once? Does their size change once the model is set up? Kokkos is using reference counting to track the memory for these views, which are created for each constraint in each convergence iteration of the generalized alpha solve. Is there a way to make this more efficient?
> 
> Why is a view being created to pass the quaternion components to create a rotation matrix? This seems like unnecessary overhead when `EulerParameterToRotationMatrix` could take 4 arguments.
> 
> Also, using `parallel_for` for 3 elements seems inefficient. I expect there will overhead to running parallel operations which won't be negligible unless the number of elements is much larger.
> 
> Why is the data type for a rotation matrix `Kokkos::View<double **>` instead of `Kokkos::View<double[3][3]>`?
> 
> Why create new views and copy `gen_cords(0:3)` and `position_vector` instead of using them directly when calculating `x0_u0`? If it's being done for clarity, the Kokkos documentation says that the views are effectively pointers so you could do the following and avoid allocating new views.
> 
> `auto x0 = position_vector;` `auto u0 = Kokkos::subview(gen_coords, Kokkos::make_pair(0,3));`

Thanks for your above and line-by-line review, @deslaughter - it was one of the more helpful reviews we did in this repo. The points you made regarding creating unnecessary `View`s, using `KokkosBlas` for linear algebra operations, short `parallel_for`s make sense to me and I removed such instances. I will keep removing these from the codebase as I work on different parts of the code and going forward, we should create some cleanup tasks after the MVP is done.

In addition, I changed variable names to reflect what they represent rather than the symbols - I think it's more readable this way, however open to feedback on it.

src/gebt_poc/solver.cpp
```diff
@@ -1,5 +1,11 @@
 #include "src/gebt_poc/solver.h"
 
+#include <KokkosBlas1_axpby.hpp>
```
deslaughter, line 3: All the blas includes can be reduced to `#include <KokkosBlas.hpp>`

deslaughter (approved):
This looks much better with fewer views created in the functions. I expect that a structure will need to be created which holds temporary views that can be reused for various operations (rotation matrix creation/multiplication, 3x1 vector math) to avoid creating any views. Then these functions could be methods on that structure.

faisal-bhuiyan:
> This looks much better with fewer views created in the functions. I expect that a structure will need to be created which holds temporary views that can be reused for various operations (rotation matrix creation/multiplication, 3x1 vector math) to avoid creating any views. Then these functions could be methods on that structure.

Sounds like a reasonable plan to me - let's keep an eye on this improvement and implement in a future work as appropriate with the MVP goals.

ddement (approved):

ddement:
> This looks much better with fewer views created in the functions. I expect that a structure will need to be created which holds temporary views that can be reused for various operations (rotation matrix creation/multiplication, 3x1 vector math) to avoid creating any views. Then these functions could be methods on that structure.

I anticipate that we'll eventually use team-based parallelism and use team scratch memory for a lot of the small working matrices.  The other alternative is to use locally defined stack arrays for some of this stuff (defined within a parallel_for statement, for example), since the dimensions for these structures are pretty small and known at compile time.

PR #88, opened 2023-11-13
Refactor static beam solver functions

faisal-bhuiyan:
This is a refactoring pull request (i.e. no new unit test added) to clean up the static beam solver functions. The goal was to improve usage of `Kokkos` data structures in the code, specifically to

- Remove allocation of unnecessary `View`s
- Use `KokkosBlas` for linear algebra operations (as opposed to using our own definitions)
- Pass `View`s into functions to get return values rather than dynamically creating them etc.

It should be helpful for our next step of solver development where we pull these functions into a linearization parameters class.

### DIFF
--- a/src/gebt_poc/solver.cpp
+++ b/src/gebt_poc/solver.cpp
@@ -317,90 +317,47 @@ void CalculateQMatrix(
     KokkosBlas::gemm("T", "N", 1., values, val, 0., non_zero_term);
 }
 
-Kokkos::View<double**> CalculateIterationMatrixComponents(
+void CalculateIterationMatrixComponents(
     const Kokkos::View<double*> elastic_force_fc, const Kokkos::View<double*> pos_vector_derivatives,
     const Kokkos::View<double*> gen_coords_derivatives,
-    const Kokkos::View<double**> sectional_stiffness
+    const Kokkos::View<double**> sectional_stiffness, Kokkos::View<double**> O_P_Q_matrices
 ) {
-    // N consists of first three components of the fc vector
-    auto N = Kokkos::View<double*>("N", kNumberOfVectorComponents);
-    Kokkos::parallel_for(
-        kNumberOfVectorComponents, KOKKOS_LAMBDA(const size_t i) { N(i) = elastic_force_fc(i); }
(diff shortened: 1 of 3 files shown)


ddement:
A quick comment before I review: when you know the size of a matrix at compile time (for example, a 6x6 stiffness matrix, or the generalized coordinates vector), you can define the view as, for example `Kokkos::View<double[6][6]>`.  If you do this for a function argument, (ie `void myFunction(Kokkos::View<double[6][6]> result)` it will provide runtime bounds checking automatically.

faisal-bhuiyan:
> A quick comment before I review: when you know the size of a matrix at compile time (for example, a 6x6 stiffness matrix, or the generalized coordinates vector), you can define the view as, for example `Kokkos::View<double[6][6]>`. If you do this for a function argument, (ie `void myFunction(Kokkos::View<double[6][6]> result)` it will provide runtime bounds checking automatically.

Thanks for the helpful tip, I agree - I added some of the fixed sizes in the function bodies and will add them in the function signatures as well.

src/gebt_poc/solver.cpp
```diff
@@ -32,116 +32,126 @@ Kokkos::View<double*> Interpolate(
...
-    return sectional_stiffness;
 }
 
-Kokkos::View<double*> CalculateElasticForces(
+void CalculateElasticForces(
     const Kokkos::View<double*> sectional_strain, Kokkos::View<double**> rotation,
     const Kokkos::View<double*> pos_vector_derivatives,
     const Kokkos::View<double*> gen_coords_derivatives,
-    const Kokkos::View<double**> sectional_stiffness
+    const Kokkos::View<double**> sectional_stiffness, Kokkos::View<double*> elastic_forces
```
ddement, line 106: Let's return the elastic forces as 2 views here, since that's how they're both assembled and used.

ddement (approved):

deslaughter (approved):

src/gebt_poc/solver.cpp
faisal-bhuiyan: Thanks for pointing out this and the following interface improvement, I had this in mind as well.

PR #89, opened 2023-11-16
Update State class to use 2-D Kokkos views

faisal-bhuiyan:
TBD

### DIFF
--- a/src/gen_alpha_poc/generalized_alpha_time_integrator.cpp
+++ b/src/gen_alpha_poc/generalized_alpha_time_integrator.cpp
@@ -1,332 +1,330 @@
-#include "src/gen_alpha_poc/generalized_alpha_time_integrator.h"
+// #include "src/gen_alpha_poc/generalized_alpha_time_integrator.h"
 
-#include "src/gen_alpha_poc/heavy_top.h"
-#include "src/gen_alpha_poc/quaternion.h"
-#include "src/gen_alpha_poc/solver.h"
-#include "src/gen_alpha_poc/vector.h"
-#include "src/utilities/log.h"
+// #include "src/gen_alpha_poc/heavy_top.h"
+// #include "src/gen_alpha_poc/quaternion.h"
+// #include "src/gen_alpha_poc/solver.h"
+// #include "src/gen_alpha_poc/vector.h"
+// #include "src/utilities/log.h"
 
-namespace openturbine::gen_alpha_solver {
+// namespace openturbine::gen_alpha_solver {
 
-GeneralizedAlphaTimeIntegrator::GeneralizedAlphaTimeIntegrator(
-    double alpha_f, double alpha_m, double beta, double gamma, TimeStepper time_stepper,
-    bool precondition
-)
-    : kAlphaF_(alpha_f),
-      kAlphaM_(alpha_m),
-      kBeta_(beta),
-      kGamma_(gamma),
-      time_stepper_(std::move(time_stepper)),
-      is_preconditioned_(precondition) {
-    if (this->kAlphaF_ < 0 || this->kAlphaF_ > 1) {
-        throw std::invalid_argument("Invalid value for alpha_f");
-    }
+// GeneralizedAlphaTimeIntegrator::GeneralizedAlphaTimeIntegrator(
+//     double alpha_f, double alpha_m, double beta, double gamma, TimeStepper time_stepper,
+//     bool precondition
+// )
+//     : kAlphaF_(alpha_f),
+//       kAlphaM_(alpha_m),
+//       kBeta_(beta),
+//       kGamma_(gamma),
+//       time_stepper_(std::move(time_stepper)),
+//       is_preconditioned_(precondition) {
+//     if (this->kAlphaF_ < 0 || this->kAlphaF_ > 1) {
+//         throw std::invalid_argument("Invalid value provided for alpha_f");
+//     }
+//     if (this->kAlphaM_ < 0 || this->kAlphaM_ > 1) {
+//         throw std::invalid_argument("Invalid value provided for alpha_m");
+//     }
+//     if (this->kBeta_ < 0 || this->kBeta_ > 0.50) {
+//         throw std::invalid_argument("Invalid value provided for beta");
+//     }
+//     if (this->kGamma_ < 0 || this->kGamma_ > 1) {
+//         throw std::invalid_argument("Invalid value provided for gamma");
+//     }
+//     this->is_converged_ = false;
+// }
 
-    if (this->kAlphaM_ < 0 || this->kAlphaM_ > 1) {
-        throw std::invalid_argument("Invalid value for alpha_m");
-    }
+// std::vector<State> GeneralizedAlphaTimeIntegrator::Integrate(
+//     const State& initial_state, size_t n_constraints,
+//     std::shared_ptr<LinearizationParameters> linearization_parameters
+// ) {
+//     auto n_gen_coords = initial_state.GetGeneralizedCoordinates().size();
+//     auto n_velocities = initial_state.GetVelocity().size();
+//     auto n_accelerations = initial_state.GetAcceleration().size();
+//     auto n_algo_accelerations = initial_state.GetAlgorithmicAcceleration().size();
 
-    if (this->kBeta_ < 0 || this->kBeta_ > 0.50) {
-        throw std::invalid_argument("Invalid value for beta");
-    }
+//     if (n_velocities != kNumberOfLieAlgebraComponents ||
+//         n_accelerations != kNumberOfLieAlgebraComponents ||
+//         n_algo_accelerations != kNumberOfLieAlgebraComponents) {
+//         throw std::invalid_argument(
+//             "The number of velocities, accelerations, and algorithmic accelerations in the "
+//             "initial state must be of size 6 or its multiple for the lie group based generalized "
+//             "alpha integrator"
+//         );
+//     }
 
-    if (this->kGamma_ < 0 || this->kGamma_ > 1) {
-        throw std::invalid_argument("Invalid value for gamma");
-    }
+//     if (n_gen_coords != kNumberOfLieGroupComponents) {
+//         throw std::invalid_argument(
+//             "The number of generalized coordinates in the initial state must be of size "
+//             "7 or its multiple for the lie group based generalized alpha integrator"
+//         );
+//     }
 
-    this->is_converged_ = false;
-}
+//     auto log = util::Log::Get();
+//     std::vector<State> states{initial_state};
+//     auto n_steps = this->time_stepper_.GetNumberOfSteps();
+//     for (size_t i = 0; i < n_steps; i++) {
+//         this->time_stepper_.AdvanceTimeStep();
+//         auto input_state = State(
+//             states[i].GetGeneralizedCoordinates(), states[i].GetVelocity(),
+//             states[i].GetAcceleration(), states[i].GetAlgorithmicAcceleration()
+//         );
+//         log->Info("** Integrating step number " + std::to_string(i + 1) + " **\n");
+//         states.emplace_back(
+//             std::get<0>(this->AlphaStep(input_state, n_constraints, linearization_parameters))
+//         );
+//     }
+//     log->Info("Time integration has completed!\n");
+//     return states;
+// }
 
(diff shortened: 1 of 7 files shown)


src/gen_alpha_poc/state.h
```diff
@@ -7,37 +7,51 @@ namespace openturbine::gen_alpha_solver {
...
+        Kokkos::View<double* [kNumberOfLieGroupComponents]> gen_coords,
+        Kokkos::View<double* [kNumberOfLieAlgebraComponents]> velocity,
+        Kokkos::View<double* [kNumberOfLieAlgebraComponents]> accln,
+        Kokkos::View<double* [kNumberOfLieAlgebraComponents]> algo_accln
     );
 
-    /// Returns the generalized coordinates vector
-    inline Kokkos::View<double*> GetGeneralizedCoordinates() const { return generalized_coords_; }
+    /// Returns the generalized coordinates vector (read only)
+    inline Kokkos::View<const double* [kNumberOfLieGroupComponents]> GetGeneralizedCoordinates(
```
ddement, line 25: Do we want the view type to be `const double`?  Don't we want to be able to edit the data that we get through this API?

deslaughter (commented):
Is gebt_poc using the time integrator from the heavy top problem? I had figured the integrator would be copied into the gebt_poc namespace since it will need to be modified for the static problem. 

As far as the modifications to make the State class use 2D views, I think this can be merged.

deslaughter (approved):
Oops, I meant to approve this PR.

faisal-bhuiyan:
> Is gebt_poc using the time integrator from the heavy top problem? I had figured the integrator would be copied into the gebt_poc namespace since it will need to be modified for the static problem.

Yup, that's what I mentioned in my [original comment](https://github.com/Exawind/openturbine/pull/89#issue-1997083471). It was faster for me to make these changes directly into generalized alpha namespace - I will move things over to the GEBT namespace before merging this.

src/gen_alpha_poc/generalized_alpha_time_integrator.cpp
```diff
@@ -91,52 +65,67 @@ std::tuple<State, Kokkos::View<double*>> GeneralizedAlphaTimeIntegrator::AlphaSt
...
-    // Perform Newton-Raphson iterations to update nonlinear part of generalized-alpha algorithm
+    // Loop over all nodes in the system and update the generalized coordinates, velocities,
+    // accelerations, and algorithmic accelerations
     auto log = util::Log::Get();
     log->Info(
         "Performing Newton-Raphson iterations to update solution using the generalized-alpha "
         "algorithm\n"
     );
 
+    for (size_t node = 0; node < n_nodes; node++) {
```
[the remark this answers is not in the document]
faisal-bhuiyan: Thanks for the suggestion - I'll keep this comment unresolved and revisit after the POC is done.

src/gen_alpha_poc/generalized_alpha_time_integrator.cpp
```diff
@@ -91,52 +65,67 @@ std::tuple<State, Kokkos::View<double*>> GeneralizedAlphaTimeIntegrator::AlphaSt
...
+    auto gen_coords_next =
+        Kokkos::View<double**>("gen_coords_next", n_nodes, kNumberOfLieGroupComponents);
+    auto velocity_next =
+        Kokkos::View<double**>("velocity_next", n_nodes, kNumberOfLieAlgebraComponents);
+    auto acceleration_next =
+        Kokkos::View<double**>("acceleration_next", n_nodes, kNumberOfLieAlgebraComponents);
+    auto algo_acceleration_next =
+        Kokkos::View<double**>("algo_acceleration_next", n_nodes, kNumberOfLieAlgebraComponents);
+    auto delta_gen_coords =
+        Kokkos::View<double**>("delta_gen_coords", n_nodes, kNumberOfLieAlgebraComponents);
```
ddement, line 88: The second extent on these can be compile time, like we've done elsewhere

src/gen_alpha_poc/generalized_alpha_time_integrator.cpp
```diff
@@ -226,38 +216,43 @@ std::tuple<State, Kokkos::View<double*>> GeneralizedAlphaTimeIntegrator::AlphaSt
...
+
+            // Update algorithmic acceleration once Newton-Raphson iterations have ended
+            Kokkos::parallel_for(
+                size,
+                KOKKOS_LAMBDA(const size_t i) {
+                    algo_acceleration_next(node, i) +=
+                        (1. - kAlphaFLocal) / (1. - kAlphaMLocal) * acceleration_next(node, i);
+                }
+            );
+        }
```
ddement, line 247: Similar to Derek's above comment, we can probably optimize this by using an mdrange and atomic access traits.  I am leaving this comment more as a reminder to future us rather than as an immediate TODO

src/gen_alpha_poc/generalized_alpha_time_integrator.cpp
```diff
@@ -272,44 +267,49 @@ std::tuple<State, Kokkos::View<double*>> GeneralizedAlphaTimeIntegrator::AlphaSt
...
 
-    auto update_generalized_coordinates = KOKKOS_LAMBDA(size_t) {
+    // Loop over all nodes in the system and update the generalized coordinates
+    for (size_t node = 0; node < gen_coords.extent(0); node++) {
         // {gen_coords_next} = {gen_coords} + h * {delta_gen_coords}
         //
         // Step 1: R^3 update, done with vector addition
-        auto current_position = Vector{gen_coords(0), gen_coords(1), gen_coords(2)};
+        auto current_position =
+            Vector{gen_coords(node, 0), gen_coords(node, 1), gen_coords(node, 2)};
```
ddement, line 287: We can't access `gen_coords` on host (it lives in device memory).  This existing error is fixed in my run on cuda pull request, so we'll need to resolve this conflicting change in a similar manner to how I did it over here.

ddement (approved):
Looks Good.  I see some immediate conflicts with my CUDA changes that we'll have to resolve.  Would it make sense for us to merge this PR now, and I'll resolve them before merging that PR?

faisal-bhuiyan:
> Looks Good. I see some immediate conflicts with my CUDA changes that we'll have to resolve. Would it make sense for us to merge this PR now, and I'll resolve them before merging that PR?

Let's do that please.

src/gen_alpha_poc/state.cpp
```diff
@@ -3,24 +3,41 @@
...
 
 State::State()
-    : generalized_coords_("generalized_coordinates", 1),
-      velocity_("velocities", 1),
-      acceleration_("accelerations", 1),
-      algorithmic_acceleration_("algorithmic_accelerations", 1) {
+    : generalized_coords_("generalized_coordinates", 1, kNumberOfLieGroupComponents),
+      velocity_("velocities", 1, kNumberOfLieAlgebraComponents),
+      acceleration_("accelerations", 1, kNumberOfLieAlgebraComponents),
+      algorithmic_acceleration_("algorithmic_accelerations", 1, kNumberOfLieAlgebraComponents) {
```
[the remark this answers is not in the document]
faisal-bhuiyan: Makes sense, thanks for catching this.

src/gen_alpha_poc/generalized_alpha_time_integrator.cpp
faisal-bhuiyan: Sounds good, I'll leave this comment unresolved as well and revisit this PR to create an issue later.

src/gen_alpha_poc/generalized_alpha_time_integrator.cpp
ddement: I seem to have been confusing this with other, similar changes I made.  The code, as it is on main, works.  However, the new code will not run on device, because you moved this code outside of a `KOKKOS_LAMBDA` (line 284 in the old code).  

The easiest way to fix it is to revert back to calling all of this inside a `KOKKOS_LAMBDA`.  Just change your indexing technique to the new approach and then have the parallel_for go over `gen_coords.extent(0)` instead of `1`.

faisal-bhuiyan:
@ddement @deslaughter I moved over the time integrator logic to `gebt_poc` namespace. I needed to comment out the unit tests in [test_gebt_generalized_alpha_solver.cpp](https://github.com/Exawind/openturbine/blob/cfb178bf6a4ef8a5a96d0a5937c49d9298b06773/tests/unit_tests/gebt_poc/test_gebt_generalized_alpha_solver.cpp) since they were failing - I don't think it's critical to figure out why for this current proof-of-concept work. Let's do a quick spot check on this one to make sure we all agree on the changes made before I merge it.

src/gebt_poc/gen_alpha_2D.cpp
```diff
@@ -0,0 +1,334 @@
...
+        }
+
+        for (size_t node = 0; node < n_nodes; node++) {
+            // Update the velocity, acceleration, and constraints based on the increments
+            Kokkos::parallel_for(
+                size,
+                KOKKOS_LAMBDA(const size_t i) {
+                    delta_gen_coords(node, i) += delta_x(i) / h;
+                    velocity_next(node, i) += kGammaPrime * delta_x(i);
+                    acceleration_next(node, i) += kBetaPrime * delta_x(i);
```
ddement, line 235: @faisal-bhuiyan 
Is this correct?  `size` is set to `velocity.size()` which is the same as `n_nodes * kNumberOfLieAlgebraComponents`.  

Instead, line 231 should be over `kNumberOfLieAlgebraComponents`, and you need to do the correct index math to access the elements of `delta_x`.  

Am I missing something here?

faisal-bhuiyan: `velocity.size()` = 6 in this context. However, I fixed all the index related issues in my local branch and planning to push these changes once we get through this and likely a couple other PRs. Does that sound good?

ddement: As long as the results are correct, sounds good.

src/gebt_poc/gen_alpha_2D.cpp
```diff
@@ -0,0 +1,334 @@
...
+            gen_coords_next(node, 4) = q.GetXComponent();
+            gen_coords_next(node, 5) = q.GetYComponent();
+            gen_coords_next(node, 6) = q.GetZComponent();
+        };
+        Kokkos::parallel_for(1, update_generalized_coordinates);
+    }
+}
+
+bool GeneralizedAlphaTimeIntegrator::CheckConvergence(const Kokkos::View<double*> residual) {
+    // L2 norm of the residual vector should be very small (< epsilon) for the solution
```
deslaughter, line 318: The L2 norm of the residual can be calculated as `KokkosBlas::nrm2(residual)` just to save some code. May consider eliminating the function since it's really easy to check.

src/gebt_poc/gen_alpha_2D.cpp
```diff
@@ -0,0 +1,334 @@
...
+        // Perform the linear update part of the generalized alpha algorithm
+        // Algorithm from Table 1, Brüls, Cardona, and Arnold 2012
+        Kokkos::parallel_for(
+            size,
+            KOKKOS_LAMBDA(const size_t i) {
+                algo_acceleration_next(node, i) = (kAlphaFLocal * acceleration(node, i) -
+                                                   kAlphaMLocal * algo_acceleration(node, i)) /
+                                                  (1. - kAlphaMLocal);
+
+                delta_gen_coords(node, i) = velocity(node, i) +
```
deslaughter, line 109: Why is `h` not included here, but then applied during `UpdateGeneralizedCoordinates`? I know that it's equivalent, I'd just like to understand why it's done this way.

deslaughter (approved):

src/gebt_poc/gen_alpha_2D.cpp
ddement: I actually like the separate function, to improve readability.  I was planning to refactor this function at some point to
```
return KokkosBlas::nrm2(residual) < kConvergenceTolerance;
```

src/gebt_poc/gen_alpha_2D.cpp
faisal-bhuiyan: Allow me to get back to you on this please.

deslaughter: Please disregard, I found the explanation in the paper describing the algorithm.

PR #96, opened 2023-11-28
Static beam linearization parameters

faisal-bhuiyan:
TBD

### DIFF
--- a/src/gebt_poc/quadrature.h
+++ b/src/gebt_poc/quadrature.h
@@ -1,0 +1,49 @@
+#pragma once
+
+#include "src/gebt_poc/element.h"
+
+namespace openturbine::gebt_poc {
+
+/// An abstract class for providing common interface to numerical quadrature rules
+class Quadrature {
+public:
+    virtual ~Quadrature() = default;
+
+    /// Returns the number of quadrature points
+    virtual size_t GetNumberOfQuadraturePoints() const = 0;
+
+    /// Returns the quadrature points (read only)
+    virtual const std::vector<double>& GetQuadraturePoints() const = 0;
+
+    /// Returns the quadrature weights (read only)
+    virtual const std::vector<double>& GetQuadratureWeights() const = 0;
+};
+
+/// A concrete quadrature rule where the quadrature points and weights are provided by the user
+class UserDefinedQuadrature : public Quadrature {
+public:
+    UserDefinedQuadrature(
+        std::vector<double> quadrature_points, std::vector<double> quadrature_weights
+    )
+        : quadrature_points_(std::move(quadrature_points)),
+          quadrature_weights_(std::move(quadrature_weights)) {}
+
+    /// Returns the number of quadrature points
+    virtual size_t GetNumberOfQuadraturePoints() const override { return quadrature_points_.size(); }
+
+    /// Returns the quadrature points (read only)
+    virtual const std::vector<double>& GetQuadraturePoints() const override {
+        return quadrature_points_;
+    }
+
+    /// Returns the quadrature weights (read only)
+    virtual const std::vector<double>& GetQuadratureWeights() const override {
+        return quadrature_weights_;
+    }
+
+private:
+    std::vector<double> quadrature_points_;
+    std::vector<double> quadrature_weights_;
+};
+
+}  // namespace openturbine::gebt_poc
--- a/src/gebt_poc/static_beam_element.cpp
+++ b/src/gebt_poc/static_beam_element.cpp
@@ -78,16 +78,13 @@ StaticBeamLinearizationParameters::StaticBeamLinearizationParameters(
       quadrature_(quadrature) {
 }
 
-// TODO Following is a hack to make things work temporarily - we should move over to
+// TECHDEBT Following is a hack to make things work temporarily - we should move over to
 // using 2D views for the solver functions
-Kokkos::View<double*> Convert2DViewTo1DView(Kokkos::View<double**> view) {
-    auto size = view.extent(0) * view.extent(1);
-    auto result = Kokkos::View<double*>("result", size);
+void Convert2DViewTo1DView(Kokkos::View<double**> view, Kokkos::View<double*> result) {
     auto populate_result = KOKKOS_LAMBDA(size_t i) {
         result(i) = view(i / view.extent(1), i % view.extent(1));
     };
-    Kokkos::parallel_for(size, populate_result);
-    return result;
+    Kokkos::parallel_for(result.extent(0), populate_result);
 }
 
 Kokkos::View<double*> StaticBeamLinearizationParameters::ResidualVector(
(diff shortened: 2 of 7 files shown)


src/gebt_poc/static_beam_element.cpp
```diff
@@ -0,0 +1,253 @@
...
+    // quadrant_2 = Transpose([B(q)])
+    auto quadrant_2 = Kokkos::subview(
+        iteration_matrix, Kokkos::make_pair(zero, size_dofs),
+        Kokkos::make_pair(size_dofs, size_iteration)
+    );
+    auto constraints_gradient_matrix =
+        Kokkos::View<double**>("constraints_gradient_matrix", size_constraints, size_dofs);
+    ConstraintsGradientMatrix(gen_coords_1D, position_vectors_, constraints_gradient_matrix);
+    // TODO ** Question for reviewers **
+    // How to transpose a matrix using KokkosBlas?
```
ddement, line 202: A quick review says that there isn't one (maybe I'm missing something).  However, a parallel transpose can be quickly written as 
```
void transpose(Kokkos::View<const double**> in, Kokkos::View<double**> out) {
  Kokkos::parallel_for(Kokkos::MDRangePolicy< Kokkos::Rank<2> > ({0,0}, {in.extent(0), in.extent(1)}), KOKKOS_LAMBDA(int i, int j) {
    out(i, j) = in(j, i);
  });
}
```
Note that this will not work with in-place transposes.  An in-place transpose would need to be performed with swaps and ensure a square matrix.

src/gebt_poc/static_beam_element.cpp
```diff
@@ -0,0 +1,253 @@
...
+        tangent_operator(3, 3) = 1.;
+        tangent_operator(4, 4) = 1.;
+        tangent_operator(5, 5) = 1.;
+    };
+    Kokkos::parallel_for(1, populate_matrix);
+
+    auto psi_0 = Kokkos::subview(psi, Kokkos::make_pair(0, 1));
+    auto psi_1 = Kokkos::subview(psi, Kokkos::make_pair(1, 2));
+    auto psi_2 = Kokkos::subview(psi, Kokkos::make_pair(2, 3));
+    const double phi = std::sqrt(psi_0(0) * psi_0(0) + psi_1(0) * psi_1(0) + psi_2(0) * psi_2(0));
```
ddement, line 234: You can't access psi's data on host - it lives in device memory.  If psi is always of length 3, we can just use 
```
auto phi = KokkosBlas::dot(psi, psi)
```
Otherwise, make an appropriate subview and use the dot product on that.

src/gebt_poc/static_beam_element.cpp
```diff
@@ -0,0 +1,253 @@
...
+    KokkosBlas::gemm("N", "N", 1.0, constraints_gradient_matrix, tangent_operator, 0.0, quadrant_3);
+
+    return iteration_matrix;
+}
+
+void StaticBeamLinearizationParameters::TangentOperator(
+    const Kokkos::View<double*> psi, Kokkos::View<double**> tangent_operator
+) {
+    // TODO ** Question for reviewers **
+    // Any better way to create the identity matrix than parallel_for?
```
ddement, line 220: Not that I can see.  There is a capability in KokkosKernels' batched functionality, but that still requires a parallel_for call.  I'll look into us using it in the future.

src/gebt_poc/static_beam_element.cpp
```diff
@@ -0,0 +1,253 @@
...
+    auto populate_matrix = KOKKOS_LAMBDA(size_t) {
+        tangent_operator(0, 0) = 1.;
+        tangent_operator(1, 1) = 1.;
+        tangent_operator(2, 2) = 1.;
+        tangent_operator(3, 3) = 1.;
+        tangent_operator(4, 4) = 1.;
+        tangent_operator(5, 5) = 1.;
+    };
+    Kokkos::parallel_for(1, populate_matrix);
```
ddement, line 229: Should we set tangent_operator to 0 first, to ensure that the off diagonal entries are correct?

src/gebt_poc/static_beam_element.cpp
```diff
@@ -0,0 +1,253 @@
...
+            "N", "N", 1.0, psi_cross_prod_matrix, psi_cross_prod_matrix, 0.0, psi_times_psi
+        );
+
+        auto quadrant4 =
+            Kokkos::subview(tangent_operator, Kokkos::make_pair(3, 6), Kokkos::make_pair(3, 6));
+        auto temp = Kokkos::View<double**>("temp", 3, 3);
+        KokkosBlas::scal(temp, (std::cos(phi) - 1.0) / (phi * phi), psi_cross_prod_matrix);
+        KokkosBlas::axpy(1.0, temp, quadrant4);
+        KokkosBlas::scal(temp, (1.0 - std::sin(phi) / phi) / (phi * phi), psi_times_psi);
+        KokkosBlas::axpy(1.0, temp, quadrant4);
```
ddement, line 249: If we wrote this as a single parallel_for, would we still need to allocate the temp view?

src/gebt_poc/static_beam_element.cpp
deslaughter: Views are initialized with all elements 0 (https://kokkos.github.io/kokkos-core-wiki/ProgrammingGuide/View.html#initialization). If the off diagonal elements are ever set to something else, they'd need to be reinitialized.

src/gebt_poc/static_beam_element.cpp
```diff
@@ -0,0 +1,253 @@
...
+              0.2797053914892766,  // weight 2
+              0.3818300505051189,  // weight 3
+              0.4179591836734694,  // weight 4
+              0.3818300505051189,  // weight 5
+              0.2797053914892766,  // weight 6
+              0.1294849661688697   // weight 7
+          }
+      )) {
+    // Define the position vectors for the 5 node beam element
+    auto populate_position_vector = KOKKOS_LAMBDA(size_t) {
```
deslaughter, line 40: Would it make sense to have this as a 2D view to match the refactor State class?

src/gebt_poc/quadrature.h
```diff
@@ -0,0 +1,49 @@
...
+/// An abstract class for providing common interface to numerical quadrature rules
+class Quadrature {
+public:
+    virtual ~Quadrature() = default;
+
+    /// Returns the number of quadrature points
+    virtual size_t GetNumberOfQuadraturePoints() const = 0;
+
+    /// Returns the quadrature points (read only)
+    virtual const std::vector<double>& GetQuadraturePoints() const = 0;
```
deslaughter, line 16: Should the points and weights be in a view so they can be accessed during integration?

faisal-bhuiyan: Let's add this as a TODO for future, I don't think it's within the scope for this PR.

ddement (approved):

src/gebt_poc/static_beam_element.cpp
```diff
@@ -0,0 +1,245 @@
...
+        tangent_operator(1, 1) = 1.;
+        tangent_operator(2, 2) = 1.;
+        tangent_operator(3, 3) = 1.;
+        tangent_operator(4, 4) = 1.;
+        tangent_operator(5, 5) = 1.;
+    };
+    Kokkos::parallel_for(1, populate_matrix);
+
+    const double phi = KokkosBlas::nrm2(psi);
+    if (std::abs(phi) > kTolerance) {
```
deslaughter, line 224: Don't need to do `abs` here, L2 norm is always positive.

deslaughter (approved):

src/gebt_poc/static_beam_element.cpp
faisal-bhuiyan: Thanks for catching this, I meant to remove this yesterday.

deslaughter:
Here are the following changes to the CI:
- Specify version (tag `4.2.00`) of Kokkos and KokkosKernels to clone and build
- Add CMake debug and bounds checking flags to Kokkos build
    - `-DCMAKE_BUILD_TYPE=Debug`
    - `-DKokkos_ENABLE_COMPILER_WARNINGS=ON`
    - `-DKokkos_ENABLE_DEBUG=ON`
    - `-DKokkos_ENABLE_DEBUG_BOUNDS_CHECK=ON`
- Add  `-DCMAKE_BUILD_TYPE=Debug` flag to KokkosKernels build
- Use `cmake --build .` command to build libraries and executables instead of `make`
- Specify number of cores to use when building `-j 2`

PR #97, opened 2023-11-29
Fixed runtime errors when running unit tests in a CUDA build

ddement:
Most of these errors are associated with memory space access errors,
while some are more nuanced aspects of Kokkos usage.  For example,
`Kokkos::deep_copy` requires views in different memory spaces to
have contiguous storage, complicating some places where we use
subviews.

This commit also removes some usages of `std::vector` in favor of
`Kokkos::View`s in support of running efficiently on device.  

The CMake script now supports the GCC 8 compiler, which needs
some extra linker flags compared to more modern versions.  

All unit tests now run on a CUDA build of OpenTurbine.

### DIFF
--- a/src/gen_alpha_poc/quaternion.h
+++ b/src/gen_alpha_poc/quaternion.h
@@ -258,14 +258,18 @@ inline Kokkos::View<double**> EulerParameterToRotationMatrix(const Kokkos::View<
     auto tilde_c_tilde_c = gen_alpha_solver::multiply_matrix_with_matrix(tilde_c, tilde_c);
 
     auto rotation_matrix = Kokkos::View<double[3][3]>("rotation_matrix");
-    Kokkos::parallel_for(1, KOKKOS_LAMBDA(std::size_t) {
-      for(std::size_t i = 0; i < 3; ++i) {
-        for(std::size_t j = 0; j < 3; ++j) {
-            rotation_matrix(i, j) =
-                identity_matrix(i, j) + 2 * euler_param(0) * tilde_c(i, j) + 2 * tilde_c_tilde_c(i, j);
+    Kokkos::parallel_for(
+        1,
+        KOKKOS_LAMBDA(std::size_t) {
+            for (std::size_t i = 0; i < 3; ++i) {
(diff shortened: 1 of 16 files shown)


src/gen_alpha_poc/quaternion.h
```diff
@@ -308,32 +308,24 @@ Quaternion rotation_matrix_to_quaternion(const RotationMatrix& rotation_matrix)
 }
 
 /// Returns the B derivative matrix given for Euler parameters, i.e. unit quaternions
-inline Kokkos::View<double**> BMatrixForQuaternions(const Quaternion& quaternion) {
```
alanw0, line 311: I'm not clear on why you converted from return-value to in-out argument. Is it related to converting the View template parameter from `double**` to `double[3][4]` ? It seems like you could have done that either way.

tests/unit_tests/CMakeLists.txt
```diff
@@ -18,9 +18,13 @@ target_compile_options(
 target_include_directories(${oturb_unit_test_exe_name} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
 target_include_directories(${oturb_unit_test_exe_name} PRIVATE ${PROJECT_BINARY_DIR})
 
+if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "9.0")
```
alanw0, line 21: Is this specific to gcc?

src/gen_alpha_poc/quaternion.h
ddement: It could be done either way.  This was mostly a style choice to separate memory allocation from math.  The material change is changing the input argument from a `Quaternion` to a `Kokkos::View<const double[4]>` to ensure host/device memory consistency.  Since I was already changing the function signature, I went ahead and changed the output style, too.

tests/unit_tests/CMakeLists.txt
ddement: To my knowledge, yes.  There may be similar issues for other compilers, but I haven't tested them.  I should add an extra check to only add this flag for GCC.

alanw0 (approved):

PR #98, opened 2023-12-05
Quaternion re-normalization after interpolation

deslaughter:
This PR adds normalization to the quaternion part of the generalized coordinates in the `Interpolate` routine. It also creates an `InterpolateDerivative` routine which applies the Jacobian when calculating the generalized coordinates derivatives. Previously, the Jacobian was passed to `Interpolate` and was set to `1` when not calculating the derivatives. Since the Jacobian is not used for non-derivative interpolation, the argument was removed from the `Interpolate` function.

Both routines were modified to use `KokkosBlas::axpy` to more efficiently perform the interpolation calculation. In the future it would be good to refactor the interpolation to use matrix multiplication and store the Lagrange interpolating polynomial weights so they aren't recalculated.

The quaternion normalization in `Interpolate` is performed via `KokkosBlas::nrm2` and then applied to the quaternion using `KokkosBlas::scal` if the norm is nonzero.

This PR also fixes some incorrectly sized views in `CalculateStaticIterationMatrix` and `CalculateStaticResidual`.

The affected unit tests were also updated.


src/gebt_poc/solver.cpp
```diff
@@ -16,22 +16,46 @@ UserDefinedQuadrature::UserDefinedQuadrature(
 }
 
 void Interpolate(
```
ddement, line 18: Should we rename this `InterpolateQuaternion`, since it can't be used for anything else?

src/gebt_poc/solver.cpp
```diff
@@ -16,22 +16,46 @@ UserDefinedQuadrature::UserDefinedQuadrature(
...
+        auto index = i * kNumberOfLieAlgebraComponents;
+        KokkosBlas::axpy(
+            interpolation_function(i),
+            Kokkos::subview(
+                nodal_values, Kokkos::pair(index, index + kNumberOfLieAlgebraComponents)
+            ),
+            interpolated_values
         );
     }
+    KokkosBlas::scal(interpolated_values, 1. / jacobian, interpolated_values);
```
ddement, line 58: Let's put a check that the Jacobian value is non-zero, similar to the check in the other function

src/gebt_poc/solver.cpp
```diff
@@ -16,22 +16,46 @@ UserDefinedQuadrature::UserDefinedQuadrature(
 }
 
 void Interpolate(
+    Kokkos::View<double*> nodal_values, Kokkos::View<double*> interpolation_function,
+    Kokkos::View<double*> interpolated_values
+) {
+    const auto n_nodes = nodal_values.extent(0) / kNumberOfLieAlgebraComponents;
+    KokkosBlas::fill(interpolated_values, 0.);
```
ddement, line 23: We can use `KokkosBlas::axpby` instead of first filling interpolated values with 0.

src/gebt_poc/solver.cpp
```diff
@@ -16,22 +16,46 @@ UserDefinedQuadrature::UserDefinedQuadrature(
...
+    for (std::size_t i = 0; i < n_nodes; ++i) {
+        auto index = i * kNumberOfLieAlgebraComponents;
+        KokkosBlas::axpy(
+            interpolation_function(i),
```
ddement, line 27: We can't access interpolation_function on host like this.  We either need to copy interpolation_function to host or use another method to keep data on device.

src/gebt_poc/solver.cpp
```diff
@@ -16,22 +16,46 @@ UserDefinedQuadrature::UserDefinedQuadrature(
...
-            Kokkos::Sum(Kokkos::subview(interpolated_values, i))
+    KokkosBlas::fill(interpolated_values, 0.);
+    for (std::size_t i = 0; i < n_nodes; ++i) {
+        auto index = i * kNumberOfLieAlgebraComponents;
+        KokkosBlas::axpy(
+            interpolation_function(i),
+            Kokkos::subview(
+                nodal_values, Kokkos::pair(index, index + kNumberOfLieAlgebraComponents)
+            ),
+            interpolated_values
```
ddement, line 55: Similar comments to the equivalent lines in the Interpolate function

src/gebt_poc/solver.cpp
deslaughter: Both of these functions are really setup to interpolate the nodal values, how about `InterpolateNodalValues` and `InterpolateNodalValueDerivatives`?

ddement: That works for me.

src/gebt_poc/solver.cpp
deslaughter: I added a check which throws an exception. However, it would be better to check that the jacobian is nonzero where it is calculated.

src/gebt_poc/solver.cpp
deslaughter: I've changed the function signature so that the data remains on the host.

ddement (approved):

PR #99, opened 2023-12-05
Implemented linear solver based on KokkosBatched instead of KokkosBlas.

ddement:
A new interface is located in the gebt_poc namespace, while the old interface's implementation was changed to call the new one.

This implementation solves several problems. First, it allows us to have more error checking than from KokkosBlas.  Second, it removes our dependence on the Magma library.  Third, it better reflects how we will need to solve matrices in the future (many small solves rather than one big solve).

No pivoting is used currently.  KokkosBatched offers pivoting, but it fails in some cases, including for one of the heavy top regression tests.


faisal-bhuiyan:
@ddement @deslaughter  I did not go through an in depth line-by-line review for this one, however, at first glance `KokkosBatched` seems like a better alternative than `KokkosBlas`. I added some logic to the current `KokkosBlas` based solver which should get us though the milestone, so let's spend some time and make sure this linear systems solver is extremely robust and thoroughly tested for every reasonable use case we can think of. I think we should get some stability in OpenTurbine for these core capabilities.

> A new interface is located in the gebt_poc namespace, while the old interface's implementation was changed to call the new one.

We should deprecate the previous solver rather than maintain two separate interfaces.

> This implementation solves several problems. First, it allows us to have more error checking than from KokkosBlas. Second, it removes our dependence on the Magma library. Third, it better reflects how we will need to solve matrices in the future (many small solves rather than one big solve).

We need to add unit tests that showcase the error checking capabilities of `KokkosBatched::gesv`. It's concerning to me that singular matrices/NaN values in the solution are not being checked and we needed to add it.

ddement:
@faisal-bhuiyan What do you think of us going ahead and merging this in to main?  It should be completely transparent to your upcoming pull requests.

faisal-bhuiyan:
> @faisal-bhuiyan What do you think of us going ahead and merging this in to main? It should be completely transparent to your upcoming pull requests.

@ddement Your changes look good to me. I added one more check to make sure `[A] {x} - {b} = {0}`. Please take a look and make sure things look good there.

faisal-bhuiyan (approved):

deslaughter (approved):

PR #101, opened 2023-12-08
Update Constraint Residual Vector for new theory

deslaughter:
This PR updates the calculation of the constraint residual vector based on the latest revisions to the theory document. The constraint residual vector is specific to the geometrically exact beam theory proof of concept in that the prescribed displacement and rotation of the root node is zero, so the residual is simply the displacement and rotation of the first node. The rotation is represented by the rotation vector (logarithmic map) of the quaternion. The implementation for calculating a rotation vector from a quaternion (`rotation_vector_from_quaternion`) was updated to avoid loss of data due to floating point number precision. A new unit test was added to verify that the rotation vector is calculated correctly for extremely small rotation angles.

### DIFF
--- a/src/gebt_poc/solver.cpp
+++ b/src/gebt_poc/solver.cpp
@@ -469,8 +469,9 @@ void ConstraintsResidualVector(
     auto translation_0 = Kokkos::subview(gen_coords, Kokkos::make_pair(0, 3));
     auto rotation_0 = Kokkos::subview(gen_coords, Kokkos::make_pair(3, 7));
 
-    // For the GEBT, proof of concept problem, the target translation and rotation is zero,
-    // so the residual vector is based on the generalized coordinates at the first node
+    // For the GEBT proof of concept problem (i.e. the clamped beam), the dofs are enforced to be
+    // zero at the left end of the beam, so the constraint residual is simply based on the
+    // generalized coordinates at the first node
     Kokkos::parallel_for(
         1,
         KOKKOS_LAMBDA(std::size_t) {
--- a/src/gebt_poc/solver.cpp
+++ b/src/gebt_poc/solver.cpp
@@ -468,28 +468,28 @@ void ConstraintsResidualVector(
 ) {
     auto translation_0 = Kokkos::subview(gen_coords, Kokkos::make_pair(0, 3));
     auto rotation_0 = Kokkos::subview(gen_coords, Kokkos::make_pair(3, 7));
-    auto rotation_matrix_0 = gen_alpha_solver::EulerParameterToRotationMatrix(rotation_0);
-    auto position_0 = Kokkos::subview(position_vector, Kokkos::make_pair(0, 3));
 
-    // position = position_0 + translation_0
-    auto position = Kokkos::View<double[kNumberOfVectorComponents]>("position");
-    Kokkos::deep_copy(position, position_0);
-    KokkosBlas::axpy(1., translation_0, position);
-
-    // rotated_position = rotation_matrix_0 * position
-    auto rotated_position = Kokkos::View<double[kNumberOfVectorComponents]>("rotated_position");
-    KokkosBlas::gemv("N", 1., rotation_matrix_0, position, 0., rotated_position);
-
-    // Assemble the constraint residual vector
-    // {constraints_residual}_6x1 = {
-    //    {translation_0}_3x1
-    //    {rotated_position - position_0}_3x1
-    // }
-    auto constraints_residual_1 = Kokkos::subview(constraints_residual, Kokkos::make_pair(0, 3));
-    Kokkos::deep_copy(constraints_residual_1, translation_0);
-    auto constraints_residual_2 = Kokkos::subview(constraints_residual, Kokkos::make_pair(3, 6));
-    Kokkos::deep_copy(constraints_residual_2, rotated_position);
-    KokkosBlas::axpy(-1., position_0, constraints_residual_2);
+    // For the GEBT, proof of concept problem, the target translation and rotation is zero,
+    // so the residual vector is based on the generalized coordinates at the first node
+    Kokkos::parallel_for(
+        1,
+        KOKKOS_LAMBDA(std::size_t) {
+            // Construct rotation vector from root node rotation quaternion
+            auto rotation_vector = openturbine::gen_alpha_solver::rotation_vector_from_quaternion(
+                openturbine::gen_alpha_solver::Quaternion(
+                    gen_coords(3), gen_coords(4), gen_coords(5), gen_coords(6)
+                )
+            );
+            // Set residual as translation and rotation of root node
+            // TODO: update when position & rotation are prescribed
+            constraints_residual(0) = gen_coords(0);
+            constraints_residual(1) = gen_coords(1);
+            constraints_residual(2) = gen_coords(2);
+            constraints_residual(3) = rotation_vector.GetXComponent();
+            constraints_residual(4) = rotation_vector.GetYComponent();
+            constraints_residual(5) = rotation_vector.GetZComponent();
+        }
+    );
 }
 
 void ConstraintsGradientMatrix(
--- a/src/gebt_poc/solver.cpp
+++ b/src/gebt_poc/solver.cpp
@@ -466,9 +466,6 @@ void ConstraintsResidualVector(
     const Kokkos::View<double*> gen_coords, const Kokkos::View<double*> position_vector,
     Kokkos::View<double*> constraints_residual
 ) {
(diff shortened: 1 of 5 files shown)


src/gebt_poc/solver.cpp
```diff
@@ -468,28 +468,28 @@ void ConstraintsResidualVector(
 ) {
     auto translation_0 = Kokkos::subview(gen_coords, Kokkos::make_pair(0, 3));
     auto rotation_0 = Kokkos::subview(gen_coords, Kokkos::make_pair(3, 7));
```
ddement, line 470: These variables appear unused.  Either use them instead of gen_coords in the lambda or delete them here.

tests/unit_tests/gebt_poc/test_solver.cpp
```diff
@@ -625,7 +625,7 @@ TEST(SolverTest, ConstraintsResidualVector) {
     ConstraintsResidualVector(generalized_coords, position_vectors, constraints_residual);
 
     openturbine::gen_alpha_solver::tests::expect_kokkos_view_1D_equal(
-        constraints_residual, {0.1, 0., 0.12, 0.1, -0.01198, 0.1194}
+        constraints_residual, {0.1, 0., 0.12, 0.1, 0., 0.0}
```
faisal-bhuiyan, line 628: Let's consider adding a comment that we expect the `constraints_residual` to be exactly same as the `generalized_coords` where the rotation part went through a log map.


```
// constraints_residual should be same as the generalized_coords where
// q{0.9987502603949662, 0.049979169270678324, 0., 0.} -> v{0.1, 0., 0.}
constraints_residual, {0.1, 0., 0.12, 0.1, 0., 0.}
```

src/gen_alpha_poc/quaternion.h
```diff
@@ -157,17 +157,18 @@ Vector rotation_vector_from_quaternion(const Quaternion& quaternion) {
...
-    // Return the rotation vector {0, 0, 0} if provided quaternion is null
-    if (close_to(sin_angle_squared, 0.)) {
-        return Vector{0.0, 0.0, 0.0};
+    Vector n(q1, q2, q3);                           // Vector part of quaternion
+    double n_length = n.Length();                   // magnitude of vector
+    double n_length_squared = n_length * n_length;  // magnitude squared
+
+    // If the length of the vector is less than machine precision,
+    // calculate and return rotation vector using method stable for small angles
+    if (n_length < kEpsilon) {
```
ddement, line 166: Is machine precision the right value here, or do we want to do this at some multiple of kEpsilon? We may want to have some unit tests the exercise this limit to make sure we're switching methods at the right time.

tests/unit_tests/gen_alpha_poc/test_quaternions.cpp
```diff
@@ -211,6 +211,18 @@ TEST(QuaternionTest, GetRotationVectorFromNullQuaternion) {
...
+    Quaternion q(0.707107, 0., 0., 0.707107);
+    auto v = rotation_vector_from_quaternion(q);
+    auto q2 = quaternion_from_rotation_vector(v);
+    Quaternion expected(0.707107, 0., 0., 0.707107);
+
+    ASSERT_NEAR(q2.GetScalarComponent(), expected.GetScalarComponent(), 1e-6);
+    ASSERT_NEAR(q2.GetXComponent(), expected.GetXComponent(), 1e-6);
+    ASSERT_NEAR(q2.GetYComponent(), expected.GetYComponent(), 1e-6);
+    ASSERT_NEAR(q2.GetZComponent(), expected.GetZComponent(), 1e-6);
+}
```
faisal-bhuiyan, line 224: This appears to be testing exactly the same thing as the combination of following couple tests.

https://github.com/Exawind/openturbine/blob/0bcd62789576242b6a2a979a8da0908aa94ffe8e/tests/unit_tests/gen_alpha_poc/test_quaternions.cpp#L167C1-L192C1

ddement (approved):

src/gen_alpha_poc/quaternion.h
```diff
@@ -157,17 +157,18 @@ Vector rotation_vector_from_quaternion(const Quaternion& quaternion) {
...
+        return n * (2. / q0 - 2. / 3. * n_length_squared / (q0 * q0 * q0));
```
faisal-bhuiyan, line 167: I might be missing it, but do we have a unit test for this behavior?

src/gen_alpha_poc/quaternion.h
```diff
@@ -157,17 +157,18 @@ Vector rotation_vector_from_quaternion(const Quaternion& quaternion) {
...
     }
 
-    double sin_angle = std::sqrt(sin_angle_squared);
-    double k = 2. * std::atan2(sin_angle, q0) / sin_angle;
-
-    return {q1 * k, q2 * k, q3 * k};
+    // Otherwise, calculate and return rotation vector using this method
+    return n * 4. * std::atan(n_length / (q0 + std::sqrt(q0 * q0 + n_length_squared))) / n_length;
```
faisal-bhuiyan, line 171: We might be able to use a ternary operator here to make things more expressive.

```
    return (
        n_length < kEpsilon
            ? (2. / q0 - 2. / 3. * n_length_squared / (q0 * q0 * q0))
            : 4. * std::atan(n_length / (q0 + std::sqrt(q0 * q0 + n_length_squared))) / n_length
    );
```

src/gebt_poc/solver.cpp
```diff
@@ -468,28 +468,28 @@ void ConstraintsResidualVector(
...
-    //    {translation_0}_3x1
-    //    {rotated_position - position_0}_3x1
-    // }
-    auto constraints_residual_1 = Kokkos::subview(constraints_residual, Kokkos::make_pair(0, 3));
-    Kokkos::deep_copy(constraints_residual_1, translation_0);
-    auto constraints_residual_2 = Kokkos::subview(constraints_residual, Kokkos::make_pair(3, 6));
-    Kokkos::deep_copy(constraints_residual_2, rotated_position);
-    KokkosBlas::axpy(-1., position_0, constraints_residual_2);
+    // For the GEBT, proof of concept problem, the target translation and rotation is zero,
+    // so the residual vector is based on the generalized coordinates at the first node
```
faisal-bhuiyan, line 473: Apologies for being nitpicky here - would the following make things a little more clear?

```
// For the GEBT proof of concept problem (i.e. the clamped beam), the dofs are enforced to be zero at
// the left end of the beam, so the constraint residual is simply based on the generalized
// coordinates at the first node
```

faisal-bhuiyan (approved):
Looks good to me. Thanks for taking the initiative to updating the constraint residuals as well as the log map. :+1:

src/gen_alpha_poc/quaternion.h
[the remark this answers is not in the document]
deslaughter: Yes, due to the way that unit quaternions are normalized, q0 gets larger as the vector part gets smaller.

src/gen_alpha_poc/quaternion.h
deslaughter: I don't have a unit test in this code, but I did pretty significant testing in Python to verify that it was working properly. Since the issue is caused by floating point precision, machine precision seems to be correct here.

tests/unit_tests/gebt_poc/test_solver.cpp
deslaughter: Added

tests/unit_tests/gen_alpha_poc/test_quaternions.cpp
deslaughter: Removed

src/gen_alpha_poc/quaternion.h
deslaughter: Technically every time a quaternion has a zero vector, it will go through this branch. However, I've added a test case specifically for it.

src/gen_alpha_poc/quaternion.h
deslaughter: In my option, this is more difficult to read, but I changed it.

src/gebt_poc/solver.cpp
deslaughter: Updated

src/gebt_poc/solver.cpp
deslaughter: Removed

src/gen_alpha_poc/quaternion.h
faisal-bhuiyan: I think we should optimize for readability rather than minute code optimization. Honestly, we could have gone either way on this, so please feel free to revert back to original version if you want.

PR #102, opened 2023-12-09
Revise static beam formulation

faisal-bhuiyan:
This work implements the revised static beam formulation as provided in the theory companion [here](https://github.com/michaelasprague/OpenTurbineTheory/blob/main/BeamTheoryNotes/doc_static.tex).


ddement:
The two unit test failures seem to be identical to what was on the branch before my CUDA fixes were pushed.  Let me know if it looks like I changed behavior on the CPU side

faisal-bhuiyan:
@deslaughter @ddement This needs some additional work with the non-zero force regression test case, however, I think we can go ahead and merge this upon review. I'll add more regression tests (and likely logic to add generalized forces) as a separate PR.

deslaughter (approved):

PR #104, opened 2023-12-15
Refactors the generalized alpha solver to use Kokkos more efficiently.

ddement:
Changed all nested loops to use hierarchical parallelism idiomatically

Improved the use of KokkosKernels where appropriate

Decreased the number of unnecessary copies in quaternion calculations.

Preconditioner is now applied using batched formulation.


src/gebt_poc/gen_alpha_2D.cpp
```diff
@@ -99,30 +121,36 @@ std::tuple<State, Kokkos::View<double*>> GeneralizedAlphaTimeIntegrator::AlphaSt
...
                                             h * (0.5 - kBetaLocal) * algo_acceleration(node, i) +
                                             h * kBetaLocal * algo_acceleration_next(node, i);
-
+            });
+            Kokkos::parallel_for(Kokkos::TeamVectorRange(member, components), [=](std::size_t i) {
                 velocity_next(node, i) = velocity(node, i) +
                                          h * (1 - kGammaLocal) * algo_acceleration(node, i) +
                                          h * kGammaLocal * algo_acceleration_next(node, i);
-
+            });
```
alanw0, line 145: Is there a guarantee that these inner parallel_fors are not executing concurrently? I.e., does a parallel_for fence before returning? If they can start before the previous one has finished, it could create some difficult-to-find correctness issues since arrays like `algo_acceleration` get read in one loop and then written in the other.

src/gebt_poc/gen_alpha_2D.cpp
```diff
@@ -192,75 +201,58 @@ std::tuple<State, Kokkos::View<double*>> GeneralizedAlphaTimeIntegrator::AlphaSt
...
-                    KOKKOS_LAMBDA(const size_t i) {
-                        // Take negative of the solution increments to update Lagrange
-                        // multipliers
-                        delta_lagrange_mults(i) = -soln_increments(i + delta_gen_coords.size());
-                        lagrange_mults_next(i) += delta_lagrange_mults(i);
+                    Kokkos::TeamVectorRange(member, components),
+                    [=](std::size_t i) {
+                        velocity_next(node, i) += kGammaPrime * residuals(node * components + i);
+                    }
+                );
```
alanw0, line 233: why not combine these 3 inner parallel_fors into one?

ddement: TeamVectorRange should be trying to vectorize the loop under the hood (at least on CPU).  I tried to keep these loops to simple statements to give the compilier a decent chance of doing so successfully.  It also might give the CUDA warps a more regular memory access pattern... I don't have any proof that either of those statements actually make a difference in the final assembly, but they feel right.  Compiler explorer could possibly help here.

src/gebt_poc/gen_alpha_2D.cpp
ddement: I did run the unit tests a bunch using CUDA and didn't run into any problems. But that might have just been Christmas magic saving me.  I'll add some barriers here on Monday.

deslaughter (approved):

PR #105, opened 2023-12-18
Refactors the residual vector and iteration matrix to use 2D views for gen_coords and position_vector.

ddement:
The previous work used 1D views and indexing math, which is at odds with the main state data structure.


src/gebt_poc/clamped_beam.h
```diff
@@ -25,7 +25,7 @@ class ClampedBeamLinearizationParameters : public LinearizationParameters {
     /// Define a static beam element with the given position vector for the nodes, 6x6
     /// stiffness matrix, and a quadrature rule
     ClampedBeamLinearizationParameters(
-        Kokkos::View<double*> position_vectors, StiffnessMatrix stiffness_matrix,
+        Kokkos::View<double**> position_vectors, StiffnessMatrix stiffness_matrix,
```
deslaughter, line 28: Should these views have a defined size on the known dimension?

```suggestion
        Kokkos::View<double*[7]> position_vectors, StiffnessMatrix stiffness_matrix,
```

deslaughter (approved):

src/gebt_poc/clamped_beam.h
ddement: Yes

PR #109, opened 2024-02-13
Revise CI pipeline

ddement:
Improves the current CI workflow to hopefully improve ergonomics going forward.

First, the formatting checks no longer end CI if they fail, providing feedback on unit test correctness even if you forgot to run clang-format before pushing.  

Second, the correctness checks have been broken into two sections: release and debug.  release is run on both Ubuntu and MacOSX.  Debug is run on Ubuntu and includes address sanitizer and undefined behavior sanitizer to improve error detection.  The warning level has also been significantly increased for OpenTurbine with warnings now taken as errors.  

Third, checks using clang-tidy and cppcheck have been added to help identify areas for improving our usage of C++.  They pass now, but throw many warnings.  We should work to have a fully clean build (no warnings) from both of those tools.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -194,6 +194,65 @@ jobs:
           -DCMAKE_BUILD_TYPE=Debug \
           -DCMAKE_CXX_CLANG_TIDY="clang-tidy"
         cmake --build . 
+  CppCheck:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Install dependencies
+      run: sudo apt-get install libgtest-dev cppcheck 
+    - name: Cache install Kokkos
+      id: cache-kokkos
+      uses: actions/cache@v3
+      with:
+        path: ~/dependencies/kokkos
+        key: ${{runner.os}}-kokkos
+    - name: Install Kokkos
+      if: steps.cache-kokkos.outputs.cache-hit != 'true'
+      run: |
+        sudo mkdir -p ~/dependencies/kokkos
+        git clone --depth 1 --branch 4.2.00 https://github.com/kokkos/kokkos
+        cd kokkos
+        mkdir build
+        cd build
+        cmake .. \
+          -DCMAKE_INSTALL_PREFIX:PATH=~/dependencies/kokkos \
+          -DCMAKE_BUILD_TYPE=Debug \
+          -DKokkos_ENABLE_DEBUG=ON \
+          -DKokkos_ENABLE_DEBUG_BOUNDS_CHECK=ON
+        sudo cmake --build . --target install -j 2
+    - name: Cache install Kokkos-Kernels
+      id: cache-kokkos-kernels
+      uses: actions/cache@v3
+      with:
+        path: ~/dependencies/kokkos_kernels
+        key: ${{runner.os}}-kokkos-kernels
+    - name: Install Kokkos-Kernels
+      if: steps.cache-kokkos-kernels.outputs.cache-hit != 'true'
+      run: |
+        sudo mkdir -p ~/dependencies/kokkos_kernels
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -74,19 +58,210 @@ jobs:
         cd build
         cmake .. \
           -DCMAKE_INSTALL_PREFIX:PATH=~/dependencies/kokkos_kernels \
+          -DCMAKE_BUILD_TYPE=Release
+        sudo cmake --build . --target install -j 2
+    - name: Clone
+      uses: actions/checkout@v3
```
rafmudaf, line 64: Update all the `actions/checkout` lines to use v4 to address this warning:
> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

tests/unit_tests/gebt_poc/test_johnson_static_beam.h
rafmudaf, line 1: I didn't see this change explained in the PR description, so just a heads up in case it was unintentional.

.github/workflows/ci.yaml
```diff
@@ -3,29 +3,16 @@ name: OpenTurbine-CI
...
     runs-on: ${{matrix.os}}
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
         include:
           - os: ubuntu-latest
-            install_deps: sudo apt-get install liblapack-dev liblapacke-dev libgtest-dev
+            install_deps: sudo apt-get install libgtest-dev
           - os: macos-latest
-            install_deps: brew install gfortran lapack googletest
```
rafmudaf, line 28: If gfortran isn't needed here, it might be removed in line 53 as well.

tests/unit_tests/gebt_poc/test_johnson_static_beam.h
ddement: Address sanitizer detected an out of bounds access here. Some of the rows only had 35 entries instead of 36 (and one of them actually had 38). It was just 0 entries at the end of the rows, so we were getting lucky and still passing the unit tests.

rafmudaf (approved):
Overall, this is a great change. I've added some minor comments above.

The address sanitizer will be really useful. Since they're only included in the Debug job, they might not be used for regression tests as those would likely be run in Release. So just a heads up to not rely on the memory checks for those high-level tests or include them in the Release job, as well. Also FWIW the tests are very fast, at the moment, so it might be worthwhile to include the memory sanitizer in Release until speed is an issue.

tests/unit_tests/gebt_poc/test_johnson_static_beam.h
rafmudaf: Oh ok, thanks for the explanation.

.github/workflows/ci.yaml
```diff
@@ -74,19 +58,210 @@ jobs:
...
+          -DCMAKE_BUILD_TYPE=Release
+        cmake --build . -j 2
+        ./openturbine -h
+        ./openturbine_unit_tests
+  Debug:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Install dependencies
+      run: sudo apt-get install libgtest-dev 
+    - name: Cache install Kokkos
```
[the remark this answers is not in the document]
ddement: The most recent batch of commits have fixed the boiler-plate problem.  We should be able to avoid adding much in the future by using the matrix feature and disabling certain configurations to prevent combinatorial explosion.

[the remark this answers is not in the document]
rafmudaf: Ah that is great, thanks for updating

PR #110, opened 2024-02-28
Change usages of Kokkos::View<...> to use aliases

ddement:
This change is in response to some of the feedback we received in the code review to improve readability.  The various fixed and dynamic sized Views are now defined in types.hpp.  Const versions of the types are obtained by using existing definitions on the main type, e.g. View1D::const_type.  

Note another small change: I have modified how we register tests with CTest.  Before, CTest saw our unit test file as one test with no indication of which test failed, now it registers every test individually using more idiomatic GTest-CMake functions.  Running with the unit test executable directly is unchanged.  

These changes are purely cosmetic, but touch many lines of code.


deslaughter:
CppCheck has identified a several things to be fixed, do you want to take care of those before merging?

ddement:
> CppCheck has identified a several things to be fixed, do you want to take care of those before merging?

Both CppCheck and Clang-Tidy are really noisy right now.  Looking at the CppCheck output on here, none of the complaints seem to be about our code, but I may be missing some.  I'm going to tweak the settings on both of these in the near future to make them more useful and do a large-scale cleanup based on it.  Do you see any specific CppCheck complaints that you want addressed?

faisal-bhuiyan:
> This change is in response to some of the feedback we received in the code review to improve readability. The various fixed and dynamic sized Views are now defined in types.hpp. Const versions of the types are obtained by using existing definitions on the main type, e.g. View1D::const_type.

Totally agree with this change - it was much needed to collate and define the rather verbose types used in `OpenTurbine` in a single location, which makes things much more maintainable.

> Note another small change: I have modified how we register tests with CTest. Before, CTest saw our unit test file as one test with no indication of which test failed, now it registers every test individually using more idiomatic GTest-CMake functions. Running with the unit test executable directly is unchanged.

Not sure I follow the motivation here, an example would have been helpful (before and after). Anyways, this PR should have been split into two imo, just to keep the lines of change more manageable.

Issue #116, opened 2024-03-11
Make a class for GLL quadrature and use in tests


faisal-bhuiyan:
GLL quadrature is used a lot in the tests. I would just make a class for it. Seems like that is what you will use in the actual end application once you set it up as well.
https://github.com/Exawind/openturbine/blob/45aab4dbf501976640405be9250204455447dcd6/src/gebt_poc/quadrature.h#L23

_Originally posted by @psakievich in https://github.com/Exawind/openturbine/discussions/108#discussioncomment-8400321_

faisal-bhuiyan:
More context from https://github.com/Exawind/openturbine/discussions/108#discussion-6183555: There is a lot of setup in the physics problem unit tests that would be nice to abstract into functions and methods.

ddement:
@faisal-bhuiyan 
This will be addressed in my next PR simplifying the quadrature class. The GLL weights and points will come from a free function. The first version will be hard coded for 7 nodes (that's all we ever use in the tests), but we can obviously generalize it easily.

faisal-bhuiyan:
> @faisal-bhuiyan This will be addressed in my next PR simplifying the quadrature class. The GLL weights and points will come from a free function. The first version will be hard coded for 7 nodes (that's all we ever use in the tests), but we can obviously generalize it easily.

Sounds good. Let's reference this issue in the PR - it'll be a nice way to track which issues have been addressed from the code review.

faisal-bhuiyan:
Already implemented.

PR #119, opened 2024-03-12
Quadrature to views

ddement:
Changed Quadrature to just be a struct of two views, and then changed interpolation to take advantage of it now living on GPU.  The previous, BLAS-based version assumed the quadrature lived on host.  This new version is poised to take advantage of our improved parallel implementation going forward.


ddement:
Addresses #116 by moving the GLL quadrature construction into a free function

PR #129, opened 2024-03-25
Restructuring proof-of-concept

deslaughter:
This PR introduces a `restruct_poc` namespace and associated tests aimed at refactoring/restructuring the solution of the geometrically exact beam theory implementation to make better use of GPU-based parallelization. This branch is a work in progress, so this PR is created as a draft mainly for discussion and feedback.


src/restruct_poc/beams.hpp
```diff
@@ -0,0 +1,609 @@
...
+    oturb::View_Nx6x6 qp_Muu;      //
+    oturb::View_Nx6x6 qp_Cuu;      //
+    oturb::View_Nx6x6 qp_Ouu;      //
+    oturb::View_Nx6x6 qp_Puu;      //
+    oturb::View_Nx6x6 qp_Quu;      //
+    oturb::View_Nx6x6 qp_Guu;      //
+    oturb::View_Nx6x6 qp_Kuu;      //
+
+    oturb::View_NxN shape_interp;  // shape function matrix for interpolation [Nodes x QPs]
+    oturb::View_NxN shape_deriv;   // shape function matrix for derivative interp [Nodes x QPs]
```
ddement, line 298: This will be a block diagonal matrix, in general, so the dense matrix representation will have a ton of 0 multiplications.  

With the "single number of nodes/qp for all elements" structure I discussed, we could use single "nodes-per-element X qp-per-element" matrix for all elements.  We would then need one interpolation matrix for each element type when we generalize to elements with different numbers of nodes.  

With the current structure, we will need to use a sparse matrix representation here to get good performance for interpolation.  Kokkos has one that we'll need to learn how to use for the linear system anyways...

src/restruct_poc/beams.hpp
```diff
@@ -0,0 +1,609 @@
...
+
+struct Beams {
+    size_t num_beams_;  // Number of beams
+    size_t num_nodes_;  // Number of nodes
+    size_t num_qps_;    // Number of quadrature points
+
+    Kokkos::View<BeamElemIndices*> elem_indices;  // View of element node and qp indices into views
+    Kokkos::View<int*> node_state_indices;        // State row index for each node
+
+    oturb::View_3 gravity;
```
ddement, line 251: I get that gravity is the same for all nodes, but isn't this just an applied force at each node?  Does gravity need its own vector, or can it just be included there at assembly time?

src/restruct_poc/beams.hpp
```diff
@@ -0,0 +1,609 @@
...
+    oturb::View_Nx3x3 M6_3x3;  //
+    oturb::View_Nx3x3 M7_3x3;  //
+    oturb::View_Nx3x3 M8_3x3;  //
+    oturb::View_Nx3x3 M9_3x3;  //
+    oturb::View_Nx3 V1_3;      //
+    oturb::View_Nx3 V2_3;      //
+    oturb::View_Nx3 V3_3;      //
+    oturb::View_Nx4 qp_quat;   //
+
+    Beams(const BeamsInit beams_init)
```
ddement, line 317: This feels like it should be either a factory free function or a method on BeamsInit.  BeamsInit can be coupled to Beams, but Beams should not be coupled to BeamsInit, if that makes sense.

src/restruct_poc/beams.hpp
```diff
@@ -0,0 +1,609 @@
...
+#include "interpolation.h"
+#include "types.hpp"
+
+#include "src/gebt_poc/quadrature.h"
+
+namespace oturb {
+
+//------------------------------------------------------------------------------
+// Beam input data structures
+//------------------------------------------------------------------------------
```
ddement, line 18: The input data structures should be in their own file rather.

src/restruct_poc/beams_calcs.hpp
```diff
@@ -0,0 +1,760 @@
+#pragma once
```
ddement, line 1: This file really should be several files to better organize things.

ddement (commented):
I wanted to post some of my initial thoughts/questions before hacking away at this to see if you have any objections.  Do you have any local changes that might conflict with stuff I push to this branch?  I don't want us tripping over each other with conflicts.

src/restruct_poc/beams.hpp
```diff
@@ -0,0 +1,609 @@
...
+          M7_3x3("R1_3x3", num_qps),
+          M8_3x3("R1_3x3", num_qps),
+          M9_3x3("R1_3x3", num_qps),
+          V1_3("V_3", num_qps),
+          V2_3("V_3", num_qps),
+          V3_3("V_3", num_qps),
+          qp_quat("Quat_4", num_qps) {}
+
+    // Update node states (displacement, velocity, acceleration) and interpolate to quadrature points
+    void UpdateState(View_Nx7 Q, View_Nx6 V, View_Nx6 A) {
```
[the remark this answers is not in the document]
deslaughter: Sure, these can be separate functions that take the `Beams` type as a parameter.

src/restruct_poc/beams.hpp
```diff
@@ -0,0 +1,609 @@
...
+// Beams data structure
+//------------------------------------------------------------------------------
+
+struct Beams {
+    size_t num_beams_;  // Number of beams
+    size_t num_nodes_;  // Number of nodes
+    size_t num_qps_;    // Number of quadrature points
+
+    Kokkos::View<BeamElemIndices*> elem_indices;  // View of element node and qp indices into views
+    Kokkos::View<int*> node_state_indices;        // State row index for each node
```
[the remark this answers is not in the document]
deslaughter: I could see some benefits to doing this for organization. However, all the views will eventually need to be passed to the functors for performing the calculations.

src/restruct_poc/beams.hpp
deslaughter: The comment on those views is a little misleading. It should say `[Nodes x Max Element QPs]`.  The number of columns is equal to the maximum number of quadrature points in any element. Since the number of quadrature points should be pretty similar or bounded, it didn't make sense to allocate a full `[Nodes x QPs]` for all the reasons you describe. The current code avoids the sparsity at the cost of having to perform the interpolation matrix multiplication on an element by element basis. But those matrix operations will be on fully-populated dense matrices.

src/restruct_poc/beams.hpp
deslaughter: Gravity is used in the calculation of `Fg` at the quadrature points, which is based on the rotated mass matrix. Since this is part of the quadrature point calculation, and is integrated back to the nodes, I thought it made sense to have it as a view. It could also be passed into the `UpdateStates` routine, but then it'd have to be transferred to the GPU before it could be used.

src/restruct_poc/beams.hpp
deslaughter: Yeah, I'm not really happy with the way that is organized either. I have it that way because it was easy to create and initialize the `Beams` type in the test fixture. I'm open to alternatives (I'm not a C++ expert by any means and I'm probably not doing it in the most appropriate way).

src/restruct_poc/beams.hpp
deslaughter: I agree, feel free to move them or I'll take care of it at some point.

src/restruct_poc/beams_calcs.hpp
deslaughter: On the one hand, it would be easier to read if broken into smaller files; on the other, it's nice to have all the functors and required functions in one place, even if it's a little long. Then again, I'm used to 2,000-5,000 line source files :)

src/restruct_poc/beams.hpp
ddement: Right, but the structure as you have it is still square and will be sparse in general.  In the event that some elements have many quadrature points and some have few, we will have a lot of unused entries.  If we have one element with 2 nodes and one element with 10 nodes (and same number of quad points), the structure will look something like:
X X 0 0 0 0 0 0 0 0
X X 0 0 0 0 0 0 0 0
X X X X X X X X X X
X X X X X X X X X X
X X X X X X X X X X
...

That block of 0 elements gets really big when there are more low order elements or larger high order elements.  I get that you use an indexing strategy so that we don't have to do math on the 0 entries, but it will use a lot of memory.

deslaughter: I agree that it has the potential to waste an appreciable amount of memory. However, it's unlikely that there will be a significant difference in the number of quadrature points between elements. OpenFAST caps the number of quadrature points at 7 and most elements will probably need at least 3. We could record the usage percent in the data structure and see what it's like under various use cases. I'm open to an alternative representation.

src/restruct_poc/beams.hpp
ddement: Yea, I was thinking it would be a force that would be handled by our integration routine at the same level as a point load or applied aerodynamic load.  I'm not sure there's much reason to handle gravity any differently and it will give us another simple example when working out that interface.

src/restruct_poc/beams_calcs.hpp
faisal-bhuiyan: I don't mind keeping all beam related calculations in the same file, despite the rather long length. In addition to the reasons above, it makes for nice and short includes in the calling code.

We could probably improve the file name - would something like `beam_fea` make more sense?

src/restruct_poc/beams.hpp
```diff
@@ -0,0 +1,689 @@
...
+
+#include <KokkosBlas.hpp>
+
+#include "beams_calcs.hpp"
+#include "interpolation.h"
+#include "types.hpp"
+
+#include "src/gebt_poc/quadrature.h"
+
+namespace oturb {
```
faisal-bhuiyan, line 14: If we are shortening the namespace name here from `openturbine`, let's go big -- what about `ot`?

src/restruct_poc/types.hpp
```diff
@@ -0,0 +1,41 @@
...
+static constexpr double Tolerance = 1.e-16;
+
+using LieAlgebraFieldView = Kokkos::View<double* [LieAlgebraComponents]>;
+using LieGroupFieldView = Kokkos::View<double* [LieGroupComponents]>;
+using VectorFieldView = Kokkos::View<double* [VectorComponents]>;
+
+using View1D = Kokkos::View<double*>;
+using View1D_Vector = Kokkos::View<double[VectorComponents]>;
+using View1D_LieAlgebra = Kokkos::View<double[LieAlgebraComponents]>;
+using View1D_LieGroup = Kokkos::View<double[LieGroupComponents]>;
```
faisal-bhuiyan, line 20: If you're not using these types anywhere in the code, let's just go ahead and remove them.

src/restruct_poc/types.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+using View_Rot = Kokkos::View<double[3][3]>;
```
faisal-bhuiyan, line 22: I would argue for calling it by the full name - it improves the readability in my opinion and removes any confusion.

`using View_RotationMatrix = Kokkos::View<double[3][3]>;`

src/restruct_poc/types.hpp
```diff
@@ -0,0 +1,41 @@
...
+using View_3x3 = Kokkos::View<double[3][3]>;
+using View_3x4 = Kokkos::View<double[3][4]>;
+using View_6x6 = Kokkos::View<double[6][6]>;
+
+using View_N = Kokkos::View<double*>;
+using View_3 = Kokkos::View<double[3]>;
+using View_Quat = Kokkos::View<double[4]>;
```
faisal-bhuiyan, line 29: On a similar note to Line 22, could we call it `using View_Quaternion = Kokkos::View<double[4]>;`?

src/restruct_poc/types.hpp
```diff
@@ -0,0 +1,41 @@
+#pragma once
+
+#include <Kokkos_Core.hpp>
+
+namespace oturb {
+
+static constexpr std::size_t LieAlgebraComponents = 6;
+static constexpr std::size_t LieGroupComponents = 7;
+static constexpr std::size_t VectorComponents = 3;
```
faisal-bhuiyan, line 9: Let's add a prefix `k` in front of these constant per the [style guide recommendation](https://google.github.io/styleguide/cppguide.html#Constant_Names).

```
static constexpr std::size_t kLieAlgebraComponents = 6;
static constexpr std::size_t kLieGroupComponents = 7;
static constexpr std::size_t kVectorComponents = 3;
```

src/restruct_poc/beams.hpp
```diff
@@ -0,0 +1,689 @@
...
+//------------------------------------------------------------------------------
+// Beam input data structures
+//------------------------------------------------------------------------------
+
+struct BeamNode {
+    double s_;
+    std::array<double, 7> x_;
+    std::array<double, 7> q_;
+    std::array<double, 6> v_;
+    std::array<double, 6> a_;
```
faisal-bhuiyan, line 25: I would prefer to call these attributes by their full name, rather than the mathematical symbol used to represent them in the theory doc.

In addition, should try to remove using [magic numbers](https://stackoverflow.com/questions/47882/what-are-magic-numbers-and-why-do-some-consider-them-bad) if we can help it going forward. I have done this previously, but let's take this opportunity to buck the trend.

```
    std::array<double, kLieGroupComponents> position_;
    std::array<double, kLieGroupComponents> generalized_coords_; // or, displacements
    std::array<double, kLieAlgebraComponents> velocity_;
    std::array<double, kLieAlgebraComponents> acceleration_; // or, accln
```

src/restruct_poc/beams.hpp
```diff
@@ -0,0 +1,689 @@
...
+#include "src/gebt_poc/quadrature.h"
+
+namespace oturb {
+
+//------------------------------------------------------------------------------
+// Beam input data structures
+//------------------------------------------------------------------------------
+
+struct BeamNode {
+    double s_;
```
faisal-bhuiyan, line 21: Without an accompanying comment describing this attribute, it is not immediately obvious what `s_` represent in a beam node. Please see my next comment on Line 22 for more on this.

src/restruct_poc/beams_calcs.hpp
deslaughter: How about `beams_functors.hpp` since that is what the users of the header would be wanting to use. Then we could have have `beams_struct.hpp` for defining the beams data structure and `beams_functions.hpp` for functions that operate on the `Beams` structure. We'll move `Beams` initialization, `UpdateStates`, etc. to `beams_functions.hpp`.

src/restruct_poc/beams.hpp
deslaughter: We can leave the namespace as `openturbine`. When I started this I had forgotten how namespaces work and thought that I'd have to prefix every type like `openturbine::View_LieAlgebraGroup`.

src/restruct_poc/types.hpp
deslaughter: I don't think I'm using this type anymore. I'll check and remove it. Otherwise, I agree that it should have a more descriptive name.

src/restruct_poc/types.hpp
deslaughter: Yes, `View_Quaternion` is a better name and is only used in a couple places.

src/restruct_poc/beams.hpp
faisal-bhuiyan: Got it. It is kind of verbose though and I would have preferred a shorter name.

> When I started this I had forgotten how namespaces work and thought that I'd have to prefix every type like openturbine::View_LieAlgebraGroup.

Yup, that's fortunately not the case. That's why I pointed it out -- if all of our outermost code live under the same namespace (i.e. `openturbine`) we don't need to prefix it every time we call a type.

src/restruct_poc/types.hpp
faisal-bhuiyan: Thanks!

In general, I like the type names to be long, descriptive, and fully spelled out like a word. It improves the reading experience in my opinion, especially for someone who's not a vested developer of the repository.

PR #132, opened 2024-03-27
Implement rotating beam constraints

### LINKED ISSUE #120: Run rotating beam test driven via constraints
Running the rotating beam in OpenTurbine.

- Updated rotating beam branch off of `main` branch
- Implementing new constraint formulation for single beam
- Getting formulation implemented, residual vector done, working on gradient matrix

faisal-bhuiyan:
Fixes #120.


deslaughter (commented):
I think it looks pretty good. One thing to note is that X0 isn't the initial position of the constrained node, it's the difference between the constrained node's initial position and the hub node's initial position. The constraint forces the two nodes to maintain the same relative distance.

PR #133, opened 2024-04-03
Fixes errors when building on GPU.  

ddement:
The main programming issue was calling deep_copy() on GPU.  There was also an unknown linker issue related to our use of RDC.  Its nature is not fully understood, but can be fixed by using smaller kernels.

Linking still fails on solver.cpp, so this commit is just the first round of fixes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,6 @@ include(openturbine-utils)
 ########################## OPTIONS #####################################
 
 # General options for the project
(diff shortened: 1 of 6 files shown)


ddement:
I'm creating this as a PR to test on CPU rather than rebuilding my Kokkos stack (which is currently built for CUDA)

ddement:
Note: Remaining build failure is the same as in target branch.  Feel free to merge

deslaughter (approved):

PR #136, opened 2024-04-16
Break up large files into several smaller ones

ddement:
This is an attempt to make our source code only contain "one thing" per file.  This should make our code easier to navigate and the dependencies of each function easier to reason about.  It will also help us to work in parallel, since merge conflicts are less likely than with very large files.  It should also make code reviews easier, since git is less likely to get confused when displaying diffs on small files.

As a side effect, the source code on restruct_poc is now header only.  While this is probably a good best practice for performance, I don't expect it will make any real difference.  It does make editing the code easier, though, since we don't have to maintain consistency between function signatures in both the hpp and cpp files.  This may also be a good practice for delivering our code as a library for external codes (if that's the direction we choose to take), since they can just include our files instead of needing to link to us.


src/restruct_poc/CalculateMuu.hpp
deslaughter, line 1: `CalculateMuu` and `CalculateCuu` are essentially the same functor and should have been that way from the beginning. Please rename to something like `RotateSectionMatrix` and refactor to remove the duplication.

src/restruct_poc/beams.hpp
```diff
@@ -1,5 +1,175 @@
...
+          node_x0("node_x0", num_nodes_),
+          node_u("node_u", num_nodes_),
+          node_u_dot("node_u_dot", num_nodes_),
+          node_u_ddot("node_u_ddot", num_nodes_),
+          node_FE("node_force_elastic", num_nodes_),
+          node_FG("node_force_gravity", num_nodes_),
+          node_FI("node_force_inertial", num_nodes_),
+          node_FX("node_force_external", num_nodes_),
+          node_residual("node_residual", num_nodes_ * 6),
+          node_iteration("node_iteration", num_nodes_ * 6, num_nodes_ * 6),
```
deslaughter, line 126: These members need to be removed. They were originally going to be the fully assembled iteration matrix and residual vector for all beams, but the assembly uses the solver views directly.

deslaughter (commented):
I think that if we're going to have this many files, there needs to be a `beams` directory and a `solver` directory as there isn't any overlap in the functors or methods. Could we include the functors with the methods that use them since each method is in it's own file?

Do we want to change the file extensions to `.cc` to align with the style guide https://google.github.io/styleguide/cppguide.html#File_Names?

src/restruct_poc/CalculateMuu.hpp
ddement: There's quite few instances of that.  Some of our interpolation kernels should be consolidated, as well.  I was planning on a consolidation in the near future and leaving this PR to just be the "break up" step without changing any actual code.

src/restruct_poc/beams.hpp
ddement: Good idea.  Will do

ddement:
> I think that if we're going to have this many files, there needs to be a `beams` directory and a `solver` directory as there isn't any overlap in the functors or methods. 

I agree that we need some folders.

> Could we include the functors with the methods that use them since each method is in it's own file?

I considered this.  I like keeping them separate because it ensures the function objects have minimal dependencies (basically, just Kokkos), while the functions calling them are dependent on the beams and solver data structures.  Also, the function objects can get long as we add different options (see IntegrateMatrix.hpp), so this makes it easy to find the top level interface/implementation.  

> Do we want to change the file extensions to `.cc` to align with the style guide https://google.github.io/styleguide/cppguide.html#File_Names?

I don't like the '.cc' file extension, personally.  '.cpp' and '.hpp' tend to be the "modern" ones.  In general, it isn't a good idea for us to adhere strictly to the Google style guide - it was created around their specific idiosyncrasies, needs, and personal preferences at the time on writing.

src/restruct_poc/CalculateMuu.hpp
deslaughter: Sounds good, we can do it in another PR.

deslaughter (approved):
@ddement Thanks for removing the unused variables in the Beams struct. I agree with not changing the extensions, I'm not a fan of `.cc` either, just wanted to discuss since we are attempting to use the style guide. Let's add the folder when we consolidate the functors.

ddement:
@deslaughter I already put the solver stuff into its own folder, so I've added that to the PR.  That should keep some separation between the concepts.  If we want to create one for beams later, we can, as well as other consolidation work.

PR #147, opened 2024-04-22
Replace hand rolled BLAS with Kokkos-Kernels on Device

ddement:
We previously had several hand-rolled Matrix-Matrix and Matrix-Vector multiply functions.  This commit replaces them with KokkosBatched implementations.  It also removes VecScale and reorganizes some operations to improve usage of KokkosKernels.  

This commit also removes the use of global "scratch" variables and instead uses local stack memory for these temporary variables.


deslaughter (approved):
@ddement Thanks for cleaning this up, I'm sure it's better to use the KokkosBlas funcs instead.

PR #161, opened 2024-05-06
Type aliases and other minor fixes

faisal-bhuiyan:
Some quick code quality related improvements. Specifically

- Additional type aliases introduced into `types.hpp` as well as reorganizing the file
- Use of braced initialization in the code as [preferred mode of initialization](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es23-prefer-the--initializer-syntax)

### DIFF
--- a/src/restruct_poc/beams/beam_section.hpp
+++ b/src/restruct_poc/beams/beam_section.hpp
@@ -11,10 +11,7 @@ struct BeamSection {
     Array_6x6 M_star;  // Mass matrix in material frame
     Array_6x6 C_star;  // Stiffness matrix in material frame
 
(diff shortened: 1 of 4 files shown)


deslaughter (commented):
The underscores were part of the constructor argument names to prevent the shadowing errors that the CI is complaining about. I'm open to using another way to name the variables, but I agree with not allowing the shadowing.

I didn't know about the curly brace initialization, thanks for pointing that out. I'll try to do that going forward.

ddement (approved):
Changes look good.  I think the CI failures will be fixed by changes upstream?

PR #162, opened 2024-05-06
Tweak static analysis tool settings to improve their usefulness

ddement:
This commit modifies the static analysis and build configuration to improve developer productivity.  It ensures that both CppCheck and clang-tidy are identifying a useful set of code warnings that should be corrected in future commits.  Once we have a clean run from these tools, their warnings should be treated as errors going forward.  The current set of checks is preliminary and will probably have more removed or added as we work to enforce them.  

To accelerate the build/test process, code in old proof-of-concept folders is no longer built.  This code remains in the repository while we finalize our code consolidation process, but it won't be built/analyzed/tested as part of our development process.


faisal-bhuiyan:
> This commit modifies the static analysis and build configuration to improve developer productivity. 

Could you provide some before and after comparisons, just to quantify the improvements?

> It ensures that both CppCheck and clang-tidy are identifying a useful set of code warnings that should be corrected in future commits.

Could you provide some commentary on how these code warnings were selected and what are they checking for?
 
> To accelerate the build/test process, code in old proof-of-concept folders is no longer built. This code remains in the repository while we finalize our code consolidation process, but it won't be built/analyzed/tested as part of our development process.

Sounds reasonable to me.

faisal-bhuiyan (approved):

ddement:
> > This commit modifies the static analysis and build configuration to improve developer productivity.
> 
> Could you provide some before and after comparisons, just to quantify the improvements?
> 

@faisal-bhuiyan 

Before, the release build with the static checks took ~65 minutes.  Now it takes ~18 minutes.  The debug build was ~50 minutes, now it takes ~10 minutes.  Obviously, these times are noisy, but the speedup is significant.  Most, if not all, of this speed up is from removing the old code from the build/check process.

> > It ensures that both CppCheck and clang-tidy are identifying a useful set of code warnings that should be corrected in future commits.
> 
> Could you provide some commentary on how these code warnings were selected and what are they checking for?
> 

For CppCheck, we just use all of them for now - by design it isn't too noisy.

For clang-tidy, you can see the .clang-tidy file.  We enable all bugprone, cppcoreguidelines, misc, modernize, performance, portability, and readability checks.  These seem like a good set of general checks - the other groupings are specific to certain standards that don't apply to us.  There are several explicit omissions:

1.  avoid-magic-numbers: This was triggering for all of our unit tests where we have specified exact values.  Also, there are plenty of places in scientific codes like ours where "magic" numbers improve readability over giving a number a name.  For example, looping from 0 to 6 when dealing with matrices of a known size as dictated by physics.

2.  use-trailing-return-type: We don't do this anywhere, and very few codes do.  I actually do like trailing return types, but requiring them in all cases is a big stretch.  

3.  else-after-return: Having multiple return statements in a file can be good for readability, in my opinion.  This shouldn't be enforced to get a passing CI run.  

4.  identifier-length:  A set minimum variable name length sounds good, but sometimes a short one makes more sense.  For physics codes, sometimes "x" or "i" are good names.  This is especially true when the variable's scope is kept small and simple.  Good naming should be enforced through code review, since it is so subjective.

5.  non-private-member-variables-in-classes:  Sometimes public member variables are awesome - they help to keep your code short, simple, and easy to use.  Sometimes they aren't good and allow inconsistent state.  It depends on context and shouldn't be enforced by a simple rule.  

6.  include-cleaner:  This one is just very noisy when you don't include what you use directly.  While I agree with the sentiment, it was complaining about Kokkos, where you're intended to include just "Kokkos_Core.hpp" and let everything else be an implementation detail.  If we could get this to ignore TPLs, then it would be a good one to turn on.  

7.  avoid-c-arrays:  This check erroneously complains about Kokkos view declarations (i.e. `Kokkos::View<double[3][4]>`).  We should avoid using c-style arrays in general, but the noise associated with one of our key data structures makes this check inappropriate for us.

8.    assignment-in-if-condition:  I like doing assignments in if conditions.  It limits the scope of variables, which I find improves readability.  Others may disagree, but I count this as subjective, so we shouldn't automatically enforce it.  Any potential bugs that arise from typos should be caught by even a half decent testing suite.  

9.  named-parameter:  This enforces that you have a commented out name for unused parameters.  I don't like this style and prefer leaving unused parameters nameless - less visual noise.  We should also avoid having unused parameters in the first place rather than use this band-aid.

tests/unit_tests/restruct_poc/test_utilities.hpp
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#include "src/gen_alpha_poc/quaternion.h"
+#include "src/gen_alpha_poc/utilities.h"
+
+namespace openturbine::gen_alpha_solver::tests {
```
faisal-bhuiyan, line 6: We missed to update the namespace here. I will take care of this in my open PR.

PR #163, opened 2024-05-06
Unit tests for math functions in restruct_poc

faisal-bhuiyan:
Completes the work scoped in #164.

Unit tests for the functions residing in `restruct_poc/math` were added in this work.


src/restruct_poc/math/quaternion_operations.hpp
```diff
@@ -18,18 +20,20 @@ KOKKOS_INLINE_FUNCTION void QuaternionToRotationMatrix(Q q, View_Rotation R) {
...
+    v_rot(1) = 2. * (q(1) * q(2) + q(0) * q(3)) * v(0) +
                (q(0) * q(0) - q(1) * q(1) + q(2) * q(2) - q(3) * q(3)) * v(1) +
                2. * (q(2) * q(3) - q(0) * q(1)) * v(2);
-    v_rot[2] = 2. * (q(1) * q(3) - q(0) * q(2)) * v(0) + 2. * (q(2) * q(3) + q(0) * q(1)) * v(1) +
+    v_rot(2) = 2. * (q(1) * q(3) - q(0) * q(2)) * v(0) + 2. * (q(2) * q(3) + q(0) * q(1)) * v(1) +
                (q(0) * q(0) - q(1) * q(1) - q(2) * q(2) + q(3) * q(3)) * v(2);
 }
 
-inline std::array<double, 3> QuaternionRotateVector(
+// TODO This is a duplicate of the above function. We should look into merging them.
```
ddement, line 35: This (and the other array functions here) was added as a convenience function for our test code.  We use them for setting up our tests and computing initial values.  I think these functions are much better for that purpose (readability, ease of use with standard C++), while the others are better for use in our performance critical kernels.  Arrays and Views have different enough APIs and semantics that merging them doesn't make sense

src/restruct_poc/math/quaternion_operations.hpp
```diff
@@ -86,24 +94,28 @@ inline std::array<double, 4> QuaternionCompose(std::array<double, 4> q1, std::ar
...
+    const auto cos_angle = Kokkos::cos(angle / 2.);
+    const auto factor = (Kokkos::abs(angle) < 1.e-12) ? 0. : Kokkos::sin(angle / 2.) / angle;
+
     quaternion(0) = cos_angle;
-    for (int i = 0; i < 3; ++i) {
-        quaternion(i + 1) = phi(i) * factor;
-    }
+    quaternion(1) = phi(0) * factor;
+    quaternion(2) = phi(1) * factor;
+    quaternion(3) = phi(2) * factor;
```
ddement, line 115: Leave this as a loop.  It gives the compiler some hope for vectorizing the loop and communicates that those components are handled the same way.

tests/unit_tests/restruct_poc/test_math.cpp
```diff
@@ -0,0 +1,369 @@
...
+    // 90 degree rotation about the x axis
+    auto q = Kokkos::View<double[4]>("q");
+    Kokkos::parallel_for(
+        1,
+        KOKKOS_LAMBDA(const int) {
+            q(0) = 0.707107;
+            q(1) = 0.707107;
+            q(2) = 0.;
+            q(3) = 0.;
+        }
```
ddement, line 19: We can't declare KOKKOS_LAMBDAs like this in test code when building for device.  The easiest way to do this is probably
```
auto q_host = Kokkos::create_mirror(q);
q_host(0) = 0.707107;
q_host(1) = 0.707107;
q_host(2) = 0.;
q_host(3) = 0.;
Kokkos::deep_copy(q, q_host);
```
Looks like you'll need to do this in most of the tests below

tests/unit_tests/restruct_poc/test_math.cpp
```diff
@@ -0,0 +1,369 @@
+#include <gtest/gtest.h>
+
+#include "src/restruct_poc/math/quaternion_operations.hpp"
+#include "src/restruct_poc/math/vector_operations.hpp"
+#include "tests/unit_tests/gen_alpha_poc/test_utilities.h"
+
+namespace openturbine::restruct_poc::tests {
+
+TEST(QuaternionTest, ConvertQuaternionToRotationMatrix_1) {
```
ddement, line 9: Is there a better name to use here?  Possibly ConvertQuaternionToRotationMatrix_XAxisRotation?

PR #166, opened 2024-05-08
Fixes newly added unit tests which fail on GPU

ddement:
The unit tests pushed this morning segfault on GPU, because they call functions which manipulate device data on host.  This commit fixes this and consolidates some setup code.


faisal-bhuiyan (approved):

faisal-bhuiyan:
Looks like I need to find a quick way to build + run unit tests on GPU.

PR #168, opened 2024-05-14
Added a series of unit tests to cover the IntegrateMatrix kernels

ddement:


tests/unit_tests/restruct_poc/CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ target_sources(
     test_rotor.cpp
     test_math.cpp
     test_utilities.cpp
+    system/test_integrate_matrix.cpp
```
faisal-bhuiyan, line 12: Do we need to create a separate folder for this single test file?

faisal-bhuiyan (approved):
Looks good.

P.S: I pushed several commits containing minor improvements to the code from readability perspective as I was reading the code, hope it is helpful.

tests/unit_tests/restruct_poc/CMakeLists.txt
ddement: No. But my intent is for our unit test directory to have the same structure as our source directory so everything stays organized. I'll add more files in here for the tests of the other kernels, add a beams folder for those unit tests, etc. This is just step one so that we don't have to move the file later.

ddement:
> Looks good.
> 
> P.S: I pushed several commits containing minor improvements to the code from readability perspective as I was reading the code, hope it is helpful.

Your changes look good. Thanks!

PR #169, opened 2024-05-21
Add integrate elastic stiffness matrix tests

ddement:


faisal-bhuiyan (approved):
Looks good to me. I added couple of commits with some minor cleanups and consolidations -- please check to make sure they look good to you.

ddement:
> Looks good to me. I added couple of commits with some minor cleanups and consolidations -- please check to make sure they look good to you.

@faisal-bhuiyan 
I don't like those changes.  I don't care about the naming scheme, but the merging of the two helper functions is not good.  Those are two different functions with different logic and different return types.  Not only that, having such a drastic change based on an template argument makes it hard to know how to use the new function to do what you want.  Finally, the goal is to make the tests easy to read and extend (for example, to different element-index mappings), which this change makes more complex. 

Is it ok if I revert your changes and merge?

faisal-bhuiyan:
> > Looks good to me. I added couple of commits with some minor cleanups and consolidations -- please check to make sure they look good to you.
> 
> @faisal-bhuiyan I don't like those changes. I don't care about the naming scheme, but the merging of the two helper functions is not good. Those are two different functions with different logic and different return types. Not only that, having such a drastic change based on an template argument makes it hard to know how to use the new function to do what you want. Finally, the goal is to make the tests easy to read and extend (for example, to different element-index mappings), which this change makes more complex.
> 
> Is it ok if I revert your changes and merge?

Yes, go for it (but keep the name changes please). The code in these functions look too similar (almost a code smell) and I was wondering if we could merge them.

PR #177, opened 2024-05-31
Reduced memory usage by about 50%

ddement:
Removed the use of extra working matrices during system assembly, and no longer allocate unused St_left matrix for GPU solves.

Assemble system still uses Update when a simple axpy will seemingly do.  This is because axpy is about 3x slower than update.  I don't know why.


deslaughter (approved):

faisal-bhuiyan:
> Removed the use of extra working matrices during system assembly, and no longer allocate unused St_left matrix for GPU solves.

Did not look very closely at it, your explanation sounds good to me.

> Assemble system still uses Update when a simple axpy will seemingly do. This is because axpy is about 3x slower than update. I don't know why.

This is indeed surprising.

faisal-bhuiyan (approved):

PR #196, opened 2024-07-02
Add rigid constraints between nodes, create Model struct for managing nodes and constraints

deslaughter:
- This PR adds the framework for describing constraints between pairs of nodes and creates a `Model` struct with methods for creating nodes (handling node number), and adding various kinds of constraints. 
- This also fixes several downsides in the way that node initial position/orientation, displacement, velocity, and acceleration were defined. 
- A test was added to demonstrate that adding a node, without mass, can be used as a point for tying constraints together. 
- The `IEA15Rotor` test was modified to define a hub node with a prescribed displacement constraint and rigid constraints from the hub to the blade root nodes.


src/restruct_poc/model/node.hpp
```diff
@@ -0,0 +1,22 @@
...
+
+struct Node {
+    int ID;      // Node identifier
+    Array_7 x;   // Node positions and orientations
+    Array_7 u;   // Node displacement
+    Array_6 v;   // Node velocity
+    Array_6 vd;  // Node acceleration
+
+    Node(
+        int id, Array_7 position, Array_7 displacement = Array_7{0., 0., 0., 1., 0., 0., 0.},
```
faisal-bhuiyan, line 15: `id` should probably be `size_t` or `unsigned_int`.

https://stackoverflow.com/questions/131803/unsigned-int-vs-size-t

src/restruct_poc/model/node.hpp
```diff
@@ -0,0 +1,22 @@
...
+#include "src/restruct_poc/types.hpp"
+
+namespace openturbine {
+
+struct Node {
+    int ID;      // Node identifier
+    Array_7 x;   // Node positions and orientations
+    Array_7 u;   // Node displacement
+    Array_6 v;   // Node velocity
+    Array_6 vd;  // Node acceleration
```
faisal-bhuiyan, line 12: This design gives me a pause -- should this be packaged into another data structure and composed to `Node`?

src/restruct_poc/model/model.hpp
```diff
@@ -0,0 +1,59 @@
+#pragma once
+
+#include "node.hpp"
+
+#include "src/restruct_poc/solver/constraint.hpp"
+#include "src/restruct_poc/types.hpp"
+
+namespace openturbine {
+
+static Node GroundNode(-1, {0., 0., 0., 1., 0., 0., 0.});
```
faisal-bhuiyan, line 10: Adding a comment to describe the intent of a `GroundNode` might be helpful here. Also, -1 for the node number makes me a little uncomfortable -- why the special treatment?

src/restruct_poc/model/model.hpp
```diff
@@ -0,0 +1,59 @@
...
+            Constraint(this->constraints.size(), node1, node2, ConstraintType::Cylindrical)
+        );
+        return this->constraints.back();
+    }
+    const Constraint RotationControl(const Node node1, const Node node2) {
+        this->constraints.push_back(
+            Constraint(this->constraints.size(), node1, node2, ConstraintType::RotationControl)
+        );
+        return this->constraints.back();
+    }
```
faisal-bhuiyan, line 54: The interfaces of the `Add` and `Get` methods are probably warranting `private` methods (and data attributes) since the intent seems to be to go through the class objects?

src/restruct_poc/model/model.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+namespace openturbine {
+
+static Node GroundNode(-1, {0., 0., 0., 1., 0., 0., 0.});
+
+struct Model {
+    std::vector<Node> nodes;
+    std::vector<Constraint> constraints;
+
+    const Node AddNode(
```
faisal-bhuiyan, line 16: Should be return a `const Node&` from this? Also, shouldn't a corresponding `GetNode` method do this job?

src/restruct_poc/solver/constraint.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+namespace openturbine {
+
+enum ConstraintType {
+    Rigid,
+    FixedBC,
+    PrescribedBC,
+    Cylindrical,
+    RotationControl,
+};
```
faisal-bhuiyan, line 13: We should use enum class as opposed to a plain enum just to be more safe: https://stackoverflow.com/questions/18335861/why-is-enum-class-considered-safer-to-use-than-plain-enum

In addition, a little bit of description might be helpful to delineate between the types.

```
enum class ConstraintType {
    Rigid,    //< rigid body constraint
    FixedBC,
    PrescribedBC,
    Cylindrical,
    RotationControl,
};
```

faisal-bhuiyan (approved):
Did a quick review and it looks good to me. Find some minor comments in the following -- feel free to take or leave them per your discretion.

src/restruct_poc/beams/beam_node.hpp
```diff
@@ -2,16 +2,16 @@
...
-    std::array<double, 7> initial_dofs;  // Node initial positions and rotations
+    double position;  // Position of node in element on range [0, 1]
+    Node node;
 
-    BeamNode(std::array<double, 7> x) : position(0.), initial_dofs(std::move(x)) {}
-    BeamNode(double s, std::array<double, 7> x) : position(s), initial_dofs(x) {}
-    BeamNode(double s, std::array<double, 3> p, std::array<double, 4> q)
-        : position(s), initial_dofs{p[0], p[1], p[2], q[0], q[1], q[2], q[3]} {}
+    BeamNode(const Node n) : position(0.), node(n) {}
+    BeamNode(double s, Node n) : position(s), node(n) {}
```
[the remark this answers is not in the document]
deslaughter: I suppose that it wouldn't hurt since it's just getting copied on the inside.

src/restruct_poc/model/node.hpp
deslaughter: `id` is an `int` because that's the type used for indexing within the views. Otherwise, I would agree with you. I'm also using an ID of `-1` to denote an invalid node for fixed and prescribed boundary condition constraints. Which definitely needs to be documented better.

src/restruct_poc/model/node.hpp
deslaughter: Possibly. `State` contains about the same set of data, except for `x`.

src/restruct_poc/model/model.hpp
deslaughter: This should be renamed to `InvalidNode`. It is used for the unused node in `FixedBC` and `PrescribedBC`. The `-1` gets used in a bunch of the sparse index functions to know if the base node is unused.

src/restruct_poc/model/model.hpp
deslaughter: That's a good idea. The `Model` class isn't designed very well. Ideally, I'd like for this to become the interface for building the models and it would contain a vector of `BeamElement`s and `MassElement`s inputs. It's really more of a starting point so I could keep the node and constraint numbering consistent.

src/restruct_poc/model/model.hpp
deslaughter: I originally was returning a node reference, and storing it in `BeamNode`, but they become invalid once another node is added to the model. We should discuss ways to make this better and more ergonomic. I'd really like to have the ability to renumber the nodes once they've all been generated and connected to elements/constraints, but I didn't have a change to figure out how to do it yet.

src/restruct_poc/solver/constraint.hpp
deslaughter: Good call on the `enum class`. I wasn't familiar with it.

Issue #175, opened 2024-05-28
Read in YAML-based WindIO file to define a turbine


ddement:
Task: Create a class which takes in a .yaml file in the [WindIO format](https://github.com/IEAWindTask37/windIO) and parses it (probably using [yaml-cpp](https://github.com/jbeder/yaml-cpp)) to define a wind turbine.

Why:  OpenTurbine needs to be able to read an input file from disk to receive user input.  

Done means: A yaml-based WindIO input file can be read and parsed successfully (does not include creating any of the OpenTurbine data structures however).

michaelasprague:
@faisal-bhuiyan -- any updates here?

michaelasprague:
No significant process; @faisal-bhuiyan discussing idea of using python dictionary generated outside openturbine from the yaml file.

Discussing python vs yaml-cpp here ...

faisal-bhuiyan:
Working on implementing a logic to read yaml-based windIO input file using the [yaml-cpp](https://github.com/jbeder/yaml-cpp) library and validating it (in a similar way done [here](https://github.com/WISDEM/WISDEM/blob/develop/wisdem/inputs/validation.py)) using the [yaml schema validation library](https://github.com/joanvallve/yaml-schema-cpp/).

faisal-bhuiyan:
Added [yaml-cpp](https://github.com/jbeder/yaml-cpp) to OpenTurbine and added logic to read a yaml node from an input file. This task assumes a prerequisite, i.e. existence of OpenTurbine input data structures to contain the turbine information (not trivial) -- adding an issue for it.

faisal-bhuiyan:
This is 70% done -- imported the Structs based off windIO schema and added method to instantiate them. Need to wrap up this work by automating the `parse` method creation via the Python script.

PR #197, opened 2024-07-02
Parse a windIO yaml input file and instantiate the turbine structs

faisal-bhuiyan:
Completes #175.

This follows up on the work performed in #199 by adding the following
- Add yaml-cpp library to support the yaml-parsing capabilities
- Enhance the python script to add `parse()` methods to each of the mapped `Struct`s
- Use the IEA 15 MW turbine as a test case for the above


deslaughter (approved):

.github/workflows/clang-tidy.yaml
```diff
@@ -25,6 +25,17 @@ jobs:
...
+      id: cache-yaml-cpp
+      uses: actions/cache@v3
+      with:
+        path: ${{ github.workspace }}/spack
+        key: linux-yaml-cpp
+    - name: Install yaml-cpp
+      if: steps.cache-yaml-cpp.outputs.cache-hit != 'true'
+      run: |
+        source spack/share/spack/setup-env.sh
+        spack install yaml-cpp
```
ddement, line 38: Once you push this in, I'm thinking of unifying it with the action above so that we only have one "Build Dependencies" step/cache.  Any thoughts/objections?

tests/unit_tests/restruct_poc/test_yaml_parser.cpp
```diff
@@ -0,0 +1,61 @@
+#include <filesystem>
+
+#include <gtest/gtest.h>
+#include <yaml-cpp/yaml.h>
+
+#include "test_utilities.hpp"
+
+#include "src/utilities/scripts/windio_mapped_structs.hpp"
+
+namespace openturbine::restruct_poc::tests {
```
ddement, line 10: If you want to avoid the need for filesystem logic, you can store the contents of the yaml file in a string (make a test_yaml_parser.hpp with a variable `std::string IEA15File = "..."`).  Then `YAML_::Load(IEA15File)` should parse it.

src/utilities/controllers/discon_rotor_test_controller.cpp
```diff
@@ -38,8 +38,8 @@ void PITCH_CONTROLLER(
 
     // If this is the first call, output the controller input and output file names
     if (first_call) {
-        std::cout << "controller input file: " << accINFILE << std::endl;
-        std::cout << "controller output file: " << avcOUTNAME << std::endl;
+        std::cout << "controller input file: " << accINFILE << "\n";
+        std::cout << "controller output file: " << avcOUTNAME << "\n";
```
ddement, line 42: Let's go ahead and delete these lines - we probably shouldn't be writing to screen in the middle of a run like this unless we're issuing a warning of some kind.

ddement (approved):
Looks good.  My only real comment was the one on the tests to try to avoid us reading an actual file from disk when running unit tests.

faisal-bhuiyan:
@ddement Could you take a look at the errors from clang-tidy here? They are looking a little arbitrary to me since I did not see these errors the last time and I don't see them on my macOS dev environment.

ddement:
@faisal-bhuiyan 
If you look at your CI run, it actually isn't running clang-tidy on MacOS - it is emitting a warning this no ClangTidy executable was found, so it is passing because no checks are actually being done.  Please don't push that change into main.  I would actually prefer that we do our CppCheck and ClangTidy runs on all CI platforms, but didn't before when ClangTidy was taking an hour to run and we had made no attempts to make it pass.  

It is failing now because the latest main branch has now enabled enforcing clang-tidy in header files, so you can see that your new errors are from the windio-mapped-struct header.  

If you need me to, I can take a look at getting that file passing clang-tidy on Linux and push up the necessary changes.

tests/unit_tests/restruct_poc/test_yaml_parser.cpp
faisal-bhuiyan: I gave it a shot -- looks like the string literal is too long for the liking of the compiler. I'll defer to you on the best course of action here.

image

.github/workflows/clang-tidy.yaml
faisal-bhuiyan: No objections from me, feel free to take care of it when you work on the CI pipeline as discussed.

faisal-bhuiyan:
@ddement Is this good to get merged now?

tests/unit_tests/restruct_poc/test_yaml_parser.cpp
ddement: No problem.  I might play with it if I feel froggy, but it's good to go for now.

ddement:
> @ddement Is this good to get merged now?

@faisal-bhuiyan Looks good to me.  Feel free to merge and I'll get to work on the CI stuff we discussed

PR #199, opened 2024-07-02
Convert windIO schema -> C++ structs

### LINKED ISSUE #198: Convert windIO schema to C++ Structs
We need to convert the [IEA wind turbine ontology/windIO schema](https://github.com/IEAWindTask37/windIO/blob/main/windIO/turbine/IEAontology_schema.yaml) to C++ data structures. To this end, writing a quick Python script might be efficient which converts the yaml schema -> C++ `Struct`s and makes things automatic.

This task is a precursor to reading a windIO file and parsing it to create the C++ objects in OpenTurbine.

faisal-bhuiyan:
Fixes #198.

A python script was created to take in the windIO ontology schema (i.e. `IEAontology_schema.yaml`) and map it to C++ data structures (specifically, `Struct`s). The schema, python script, and the output C++ were added to the PR for completeness -- as things stand 115 unique `Struct`s were mapped from the windIO schema.

This accomplishes a couple things
- A structured way to parse windIO description of a turbine and store them into C++ Structs (we know the items we're looking for as directed by the schema)
- Any future updates to the schema can be accommodated by quickly running the script one more time

NB for reviewers: the script is only ~250 LOC which is essentially the reviewable item here.


src/utilities/scripts/windio.py
```diff
@@ -0,0 +1,256 @@
...
+
+    # print the detected type
+    #print(f"{field.name}: {field.type}")
+
+
+def main():
+    """
+    Main function to build the structs from the schema
+    """
+    with open("/Users/fbhuiyan/dev/openturbine/src/utilities/scripts/IEAontology_schema.yaml", 'r') as file:
```
ddement, line 228: This hardcodes your local path into the script.  Perhaps the input and output file names should be taken as command line data, so running the script should look something like `./windio.py -i schema.yaml -o structs.cpp`

src/utilities/scripts/windio.py
```diff
@@ -0,0 +1,256 @@
...
+    root = Schema(data)
+
+    # Build the structs from the schema
+    struct_map = {}
+    build_structs(Struct("Turbine"), root, root.definitions, struct_map)
+
+    struct_names = sorted(struct_map.keys())
+
+    # Write structs to file
+    with open("/Users/fbhuiyan/dev/openturbine/src/utilities/scripts/windio_mapped_structs.cpp", 'w') as file:
```
ddement, line 240: see commend for line 228

src/utilities/scripts/windio.py
```diff
@@ -0,0 +1,256 @@
...
+        if not schema.items:
+            raise ValueError(f"{field.name}: array without item spec")
+        field.type = schema.items.type
+        build_type(field, schema.items, definition_map, struct_map)
+        field.type = f"std::vector<{field.type}>"
+    else:
+        raise ValueError(f"Unknown type '{schema.type} - {schema}'")
+
+    # print the detected type
+    #print(f"{field.name}: {field.type}")
```
ddement, line 221: debug code?  If so, it should be removed

src/utilities/scripts/windio.py
```diff
@@ -0,0 +1,256 @@
...
+            # If the struct is the same, return
+            print(f"Struct {s.name} is the same, returning")
+            return
+
+        # If the struct is not the same, add a suffix to the name by keeping track of the number of structs with the same name
+        print(f"Struct {s.name} is not the same, adding suffix")
+        i = 1
+        while f"{s.name}_{i}" in struct_map:
+            i += 1
+        s.name = f"{s.name}_{i}"
```
ddement, line 135: Is this logic necessary for our usecase?  Do we expect IEAOntology_schema to have duplicate struct names?  If not, then we should just error out if s.name already exists.  If this is an expected use case, then let's not print to screen on 105, 127, and 131.

src/utilities/scripts/windio.py
faisal-bhuiyan: Fixed

src/utilities/scripts/windio.py
faisal-bhuiyan: > Do we expect IEAOntology_schema to have duplicate struct names?

We unfortunately need to handle this case -- the schema contains duplicate struct names with different makeup.

ddement (approved):
Looks good

deslaughter (approved):

PR #207, opened 2024-07-12
Add Cylindrical Constraint

### LINKED ISSUE #203: Add cylindrical constraint for hub rotation
The basic framework for this are in `main`. A unit test has been added to demonstrate proper functioning of the constraint.

What needs to be done:
- Add code to calculate constraint residual and constraint gradient
- Rework sparse B matrix indexing schemes to account for constraints that have fewer than 6 DOFs (rows)

deslaughter:
This PR completes the addition of the Cylindrical Constraint for modeling free hub rotation about an axis. This required modification to the constraint residual and constraint gradient calculation along with adding a method to calculate the axis of rotation for the cylindrical constraint. The routines uses to calculate the sparse indices for the constraint gradient matrix, B, were reworked because the cylindrical constraint only has 5 rows instead of the previously assumed 6 rows. Instead of recalculating the sparse indices for the transpose of B, a generic routine was created which can directly calculate the transpose indices from the non-transposed indices. This utility will likely be useful in the future as we make more use of sparse matrices.

Merging of this PR closes #203


deslaughter:
There was an issue merging `main` into this branch which created a bunch of duplicate history. Fixed it with the previous force push.

src/restruct_poc/math/vector_operations.hpp
```diff
@@ -18,12 +20,54 @@ KOKKOS_INLINE_FUNCTION void VecTilde(VectorType vector, MatrixType matrix) {
...
+    };
+}
+
+/// UnitVector returns the unit vector of the given vector
+constexpr Array_3 UnitVector(const Array_3& v) {
+    // Calculate vector norm
+    double norm = std::sqrt(v[0] * v[0] + v[1] * v[1] + v[2] * v[2]);
+    if (norm == 0) {
+        throw std::invalid_argument("Cannot normalize a zero vector");
+    }
```
ddement, line 65: Would it not be correct to just return v in this case? 

Also, I'm surprised we're not throwing a warning/error comparing a double to an integer here.

src/restruct_poc/solver/calculate_constraint_residual_gradient.hpp
```diff
@@ -194,19 +230,29 @@ struct CalculateConstraintResidualGradient {
             B(i, i) = -1.;
         }
 
-        // B12 = tilde(R1*X0)
-        VecTilde(R1_X0, A);
-        for (int i = 0; i < 3; ++i) {
+        if (cd.type == ConstraintType::kCylindrical) {
+            // B22 = cross(R1 * x0_hat, transpose(R2 * z0_hat))
+            // B22 = cross(R1 * x0_hat, transpose(R2 * y0_hat))
```
ddement, line 235: Don't these comments contradict each other?

src/restruct_poc/solver/solver.hpp
```diff
@@ -176,25 +168,22 @@ struct Solver {
...
-                this->constraints.num, this->num_system_nodes, this->constraints.data, B_t_indices}
+            "PopulateSparseRowPtrsColInds_Transpose", 1,
+            PopulateSparseRowPtrsColInds_Transpose{
+                B_num_rows, B_num_columns, col_count, tmp_row_ptrs, B_row_ptrs, B_col_ind,
+                B_t_row_ptrs, B_t_col_inds}
         );
 
         auto B_t_values = Kokkos::View<double*>("B_t values", B_t_num_non_zero);
-        KokkosSparse::sort_crs_matrix(B_t_row_ptrs, B_t_indices, B_t_values);
+        // KokkosSparse::sort_crs_matrix(B_t_row_ptrs, B_t_indices, B_t_values);
```
ddement, line 183: remove commented code

src/restruct_poc/solver/populate_sparse_row_ptrs_col_inds_transpose.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+struct PopulateSparseRowPtrsColInds_Transpose {
+    int rows;
+    int cols;
+    Kokkos::View<int*> col_count;       // cols
+    Kokkos::View<int*> temp_row_ptr;    // cols + 1
+    Kokkos::View<int*> row_ptrs;        // rows + 1
+    Kokkos::View<int*> col_inds;        // nnz
+    Kokkos::View<int*> row_ptrs_trans;  // cols + 1
+    Kokkos::View<int*> col_inds_trans;  // nnz
```
ddement, line 19: make row_ptrs and col_inds `Kokkos::View<int*>::const_type` and put them first (so that the mutable variables come last)

ddement (approved):
Looks good.  I especially like the transpose algorithm.

src/restruct_poc/math/vector_operations.hpp
deslaughter: It wouldn't be a proper unit vector without having a magnitude of 1. I think the downstream logic would fail if the vector were zero. I'm also surprised that there's no issue with the integer comparison, probably some sort of automatic type promotion. However, I can easily fix it.

src/restruct_poc/solver/calculate_constraint_residual_gradient.hpp
deslaughter: Yes. B22 is a 2x3 matrix and I was just trying to capture the two rows, but it could be expressed better

src/restruct_poc/solver/constraint.hpp
```diff
@@ -50,29 +55,33 @@ struct Constraint {
...
-                std::array<std::array<double, 3>, 3> R = {
-                    {{v[0] * v[0] * k + c, v[0] * v[1] * k - v[2], v[0] * v[2] * k + v[1]},
-                     {v[1] * v[0] * k + v[2], v[1] * v[1] * k + c, v[1] * v[2] * k - v[0]},
-                     {v[2] * v[0] * k - v[1], v[2] * v[1] * k + v[0], v[2] * v[2] * k + c}}};
-
+                // Columns of rotation matrix are orthogonal unit vectors with
+                // the first column matching the target vector
                 this->x_axis = {R[0][0], R[1][0], R[2][0]};
                 this->y_axis = {R[0][1], R[1][1], R[2][1]};
                 this->z_axis = {R[0][2], R[1][2], R[2][2]};
```
faisal-bhuiyan, line 87: Would the following make things a little more readable?

```
// If fixed BC or prescribed displacement, X0 is based on reference position vector
        if (constraint_type == ConstraintType::kFixedBC ||
            constraint_type == ConstraintType::kPrescribedBC) {
            this->X0[0] = this->target_node.x[0] - vec[0];
            this->X0[1] = this->target_node.x[1] - vec[1];
            this->X0[2] = this->target_node.x[2] - vec[2];
            return;
        }

        // Calculate initial difference in position between nodes
        this->X0[0] = this->target_node.x[0] - this->base_node.x[0];
        this->X0[1] = this->target_node.x[1] - this->base_node.x[1];
        this->X0[2] = this->target_node.x[2] - this->base_node.x[2];

        // If rotation control constraint, vec is rotation axis
        if (constraint_type == ConstraintType::kCylindrical) {
            Array_3 x{1., 0., 0.};
            auto x_hat = UnitVector(this->X0);

            // Create rotation matrix which rotates x to match vector
            auto v = CrossProduct(x, x_hat);
            auto c = DotProduct(x_hat, x);
            auto k = 1. / (1. + c);
            Array_3x3 R = {
                {{
                     v[0] * v[0] * k + c,
                     v[0] * v[1] * k - v[2],
                     v[0] * v[2] * k + v[1],
                 },
                 {
                     v[1] * v[0] * k + v[2],
                     v[1] * v[1] * k + c,
                     v[1] * v[2] * k - v[0],
                 },
                 {
                     v[2] * v[0] * k - v[1],
                     v[2] * v[1] * k + v[0],
                     v[2] * v[2] * k + c,
                 }}};

            // Columns of rotation matrix are orthogonal unit vectors with
            // the first column matching the target vector
            this->x_axis = {R[0][0], R[1][0], R[2][0]};
            this->y_axis = {R[0][1], R[1][1], R[2][1]};
            this->z_axis = {R[0][2], R[1][2], R[2][2]};
        }
    }
```

src/restruct_poc/solver/populate_sparse_row_ptrs_col_inds_constraints.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include <Kokkos_Core.hpp>
+
+#include "constraints.hpp"
+
+#include "src/restruct_poc/types.hpp"
+
+namespace openturbine {
+
+struct PopulateSparseRowPtrsColInds_Constraints {
```
faisal-bhuiyan, line 11: I like to write the full words in class names, but that's my personal taste.

`PopulateSparseRowPointersColumnIndicesForConstraints`

faisal-bhuiyan (approved):
> This PR completes the addition of the Cylindrical Constraint for modeling free hub rotation about an axis.

Would it be possible to add a quick animation of this newly added capability? I'm a sucker for these eye candies.

The PR looks good to me -- just wanted to make sure all the added functions and methods in the classes are unit tested. Please see the line-by-line comments in the following.

src/restruct_poc/solver/populate_sparse_row_ptrs_col_inds_constraints.hpp
deslaughter: I agree that these sparse kernels could be named better, possibly like `BuildConstraintSparseIndices`. I was following the existing naming. Might be a good thing to keep in mind when cleaning up since there are four such functions.

src/restruct_poc/solver/constraint.hpp
deslaughter: Yes, that is cleaner, I'll change it

src/restruct_poc/model/model.hpp
```diff
@@ -15,40 +15,40 @@ struct Model {
     std::vector<Constraint> constraints;
 
     const Node AddNode(
```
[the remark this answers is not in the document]
deslaughter: You're right, a regular node would suffice. I was thinking that it would indicate that the node values shouldn't be changed, but it doesn't really matter. Eventually I'd like for these to return a reference or pointer to the nodes in the vector so they can be manipulated, but I need some help figuring out how to do that.

src/restruct_poc/solver/populate_sparse_row_ptrs_col_inds_constraints.hpp
ddement: On one hand: yes, the naming is bad.  On the other hand, I didn't work too hard on naming because they will probably need to be reworked/rewritten a few times as we bang them into shape.  

Not only is the accuracy of the sparsity pattern for different use cases a concern, but we will also likely want the pattern to be built in parallel, which will make a difference for large problems (surprising, but I've seen it in my testing), but is non-trivial to implement.  

All of this is to say that spending too much time tweaking the naming of these kernels might not be a great use of time right now.

PR #208, opened 2024-07-16
Reduce Sparse Matrix Overhead

ddement:
Previous moves to use sparse matrix operations introduced some overhead for small problems.  

This commit moves the "symbolic" portion of all sparse matrix operations into the Solver class constructor, thus only calling them once at setup time.


ddement:
@deslaughter @faisal-bhuiyan This commit seems to run into a CI problem on MacOS.  Seeing as I'm planning on bringing over my CI rework commits next, are you OK with us going ahead and merging this now?  Perhaps one of you could try to build this branch really quickly to make sure I didn't actually break your build...

faisal-bhuiyan (approved):
Looks like a good performance improvement step to me.

> @deslaughter @faisal-bhuiyan This commit seems to run into a CI problem on MacOS. Seeing as I'm planning on bringing over my CI rework commits next, are you OK with us going ahead and merging this now? Perhaps one of you could try to build this branch really quickly to make sure I didn't actually break your build...


I was able to build this branch on macOS without any issues.

PR #209, opened 2024-07-16
Add tests and Refactor UpdateState Kernels

ddement:
In addition to unit testing, This PR provides some cleanup of the implementation of CalculateNodeForces and UpdateNodeState specifically. It also adds typedefs for the KokkosKernels types used in these kernels, which greatly improves readability.

This PR is just #204 but rebased on main.

### DIFF
--- a/src/restruct_poc/system/calculate_Ouu.hpp
+++ b/src/restruct_poc/system/calculate_Ouu.hpp
@@ -9,6 +9,9 @@
 namespace openturbine {
 
 struct CalculateOuu {
+    using NoTranspose = KokkosBatched::Trans::NoTranspose;
+    using Default = KokkosBatched::Algo::Gemm::Default;
+    using Gemm = KokkosBatched::SerialGemm<NoTranspose, NoTranspose, Default>;
     View_Nx6x6::const_type qp_Cuu_;
     View_Nx3x3::const_type x0pupSS_;
     View_Nx3x3::const_type M_tilde_;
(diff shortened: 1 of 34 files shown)


src/restruct_poc/system/calculate_Quu.hpp
```diff
@@ -20,19 +26,15 @@ struct CalculateQuu {
         auto x0pupSS = Kokkos::subview(x0pupSS_, i_qp, Kokkos::ALL, Kokkos::ALL);
         auto N_tilde = Kokkos::subview(N_tilde_, i_qp, Kokkos::ALL, Kokkos::ALL);
         auto m1 = Kokkos::Array<double, 9>{};
-        auto M1 = Kokkos::View<double[3][3], Kokkos::MemoryTraits<Kokkos::Unmanaged>>(m1.data());
+        auto M1 = Kokkos::View<double[3][3]>(m1.data());
```
faisal-bhuiyan, line 29: Looks like we could use the shorter name `View_3x3` here as defined in `src/restruct_poc/types.hpp`?

src/restruct_poc/system/calculate_Ouu.hpp
```diff
@@ -9,6 +10,9 @@
 namespace openturbine {
 
 struct CalculateOuu {
+    using NoTranspose = KokkosBatched::Trans::NoTranspose;
+    using Default = KokkosBatched::Algo::Gemm::Default;
+    using Gemm = KokkosBatched::SerialGemm<NoTranspose, NoTranspose, Default>;
     View_Nx6x6::const_type qp_Cuu_;
```
faisal-bhuiyan, line 16: Would it make sense to define these type aliases in `src/restruct_poc/types.hpp` as opposed to defining them in several places of system files?

faisal-bhuiyan (approved):
I didn't get too deep into the implementation of test code, the source code looks good to me.

src/restruct_poc/system/calculate_Quu.hpp
ddement: Yes, it seems like it.  I'll proliferate that change to the other parts of the code affected by this PR.

src/restruct_poc/system/calculate_Ouu.hpp
ddement: I like keeping aliases like this local, in case we want to do specific tweaking to the algorithm choice/configuration.  It also saves common names like these for if we want to use them differently in other contexts

PR #210, opened 2024-07-16
Refactor CI for Better Maintainability

ddement:
After several iterations of having to patch up CI, it was looking pretty messy. This clean-up simplifies our setup for each CI run (no longer depending on a github action for setting the compiler) and splits apart MacOS, Linux, ClangFormat, ClangTidy, and CppCheck into different files/jobs.

This is just #205 rebased on main.


.github/workflows/deploy-pages.yaml
faisal-bhuiyan, line 1: What did this do?

.github/workflows/correctness.yaml
faisal-bhuiyan, line 1: Should we call this `correctness-ubuntu.yaml` and the other one `correctness-macOS.yaml`?

faisal-bhuiyan (approved):
Looks good, thanks for cleaning this up. Looks much better. Maybe worth adding a README for this folder to shortly document what each of the files are doing?

.github/workflows/deploy-pages.yaml
ddement: I think it was about automatically building/deploying documentation of some kind?  It never actually did anything, as far as I can tell.

.github/workflows/correctness.yaml
ddement: Good idea.  Will do

PR #217, opened 2024-07-18
Refactor Model class to hold smart pointers as its attributes

### LINKED ISSUE #216: Refactor Model and Node classes
We need to refactor the `Model` class to store pointers to `Node` and (potentially) `Constraint` objects -- currently, we are holding copies to such objects which makes transformations difficult to keep track of.

### Change Data Structures:

- Modify the nodes vector to hold std::shared_ptr<Node>.
- Modify the constraints vector to hold std::shared_ptr<Constraint>.
- Add a std::vector<std::shared_ptr<BeamElement>> to store beam elements.

###  Update Methods to Use Pointers:

- Refactor all methods that add nodes and constraints to return pointers instead of objects.
- Ensure proper handling and storage of these pointers in the nodes, constraints, and beam_elements vectors.

###  Node Transformations:
- Add methods to the Node class to perform transformations like rotation and translation.
- Ensure transformations are reflected in the model

### Add Beam Elements:
- Implement a method to add BeamElement objects to the model.

### Testing
- Create node, add to model, transform node, make assertion on node from model class

faisal-bhuiyan:
Fixes #216.


src/restruct_poc/model/node.hpp
```diff
@@ -17,6 +18,28 @@ struct Node {
...
+        x[1] += displacement[1];
+        x[2] += displacement[2];
+    }
+
+    /// Rotate a node by a rotation axis and angle
+    void Rotate(const Array_3& axis, double angle) {
+        auto q = Array_4{
+            cos(angle / 2.), sin(angle / 2.) * axis[0], sin(angle / 2.) * axis[1],
+            sin(angle / 2.) * axis[2]};
+        auto R = QuaternionToRotationMatrix(q);
```
deslaughter, line 34: Could we use `RotateVectorByQuaternion` instead of converting to a rotation matrix? Also, I'd like to specify the point about which the node is being rotated. I think this should affect the node's rotation component as well, so you'll need to compose the node's current rotation with the quaternion that's constructed in this function. I also think this needs to modify the velocity and acceleration vectors.

Here's the test case. Create a blade in it's default orientation along the X axis with the blade root at the hub radius of 1 meters. Rotate about the blade root to give it a cone angle of 5 degrees (rotate around Y axis).  Set the rotational velocity to 5 rad/s about the Z axis (this may take some figuring out). Rotate +90 degrees around Z axis so that the blade is pointing up along +Y. Tilt the blade back to account for 5 degrees of shaft tilt (-5 degrees about X axis). Translate along +Y by 100 meters. The question is, what is the tip velocity. We may need to discuss and calculate this value. Also there may be a better order of operations, I'm open to other ideas.

faisal-bhuiyan: > Could we use RotateVectorByQuaternion instead of converting to a rotation matrix? Also, I'd like to specify the point about which the node is being rotated. I think this should affect the node's rotation component as well, so you'll need to compose the node's current rotation with the quaternion that's constructed in this function. I also think this needs to modify the velocity and acceleration vectors.

I refactored the `Rotate` method to accomplish this, however not sure what you mean by specifying the point of rotation -- it needs to be a vector and and angle right?

src/restruct_poc/model/model.hpp
```diff
@@ -15,64 +16,138 @@ static Node InvalidNode(-1, {0., 0., 0., 1., 0., 0., 0.});
...
+    std::shared_ptr<Node> AddNode(
         const Array_7& position, const Array_7& displacement = Array_7{0., 0., 0., 1., 0., 0., 0.},
         const Array_6& velocity = Array_6{0., 0., 0., 0., 0., 0.},
         const Array_6& acceleration = Array_6{0., 0., 0., 0., 0., 0.}
     ) {
-        this->nodes.push_back(Node(nodes.size(), position, displacement, velocity, acceleration));
-        return this->nodes.back();
+        auto node =
+            std::make_shared<Node>(nodes_.size(), position, displacement, velocity, acceleration);
+        this->nodes_.push_back(std::move(node));
```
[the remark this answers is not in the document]
ddement: This pattern is used in a few places.  Assume I made this comment in all of them, too.

tests/unit_tests/restruct_poc/test_model.cpp
ddement, line 1: Move this to be under a model header (like I've done for solver, system, etc).  I know it feels a little weird when it's just one file, but I like the organization of our true unit tests to be symmetrical to our source code.

src/restruct_poc/model/model.hpp
```diff
@@ -15,64 +16,138 @@ static Node InvalidNode(-1, {0., 0., 0., 1., 0., 0., 0.});
...
-    Constraint AddRigidConstraint(const Node& node1, const Node& node2) {
-        this->constraints.push_back(
-            Constraint(ConstraintType::kRigid, this->constraints.size(), node1, node2)
-        );
-        return this->constraints.back();
+    /// Return a node by ID - const/read-only version
+    std::shared_ptr<const Node> GetNode(int id) const { return this->nodes_[id]; }
+
+    /// Return a node by ID - non-const version
+    std::shared_ptr<Node> GetNode(int id) { return this->nodes_[id]; }
```
ddement, line 58: Should these return a const Node& and Node& respectively?  We never call this function where we don't immediately dereference the shared pointer, so we might as well dereference them in place.

src/restruct_poc/model/model.hpp
```diff
@@ -15,64 +16,138 @@ static Node InvalidNode(-1, {0., 0., 0., 1., 0., 0., 0.});
...
+
+    Model(
+        std::vector<std::shared_ptr<Node>> nodes,
+        std::vector<std::shared_ptr<BeamElement>> beam_elements
+    )
+        : nodes_(std::move(nodes)), beam_elements_(std::move(beam_elements)) {}
+
+    // Make the class non-copyable and non-assignable
+    Model(const Model&) = delete;
+    Model& operator=(const Model&) = delete;
```
ddement, line 40: Why do these need to be deleted?  I would propose allowing them for shallow copies and adding a "CloneModel" function (or similar name) for when we want a deep copy.

ddement (commented):
Why do we want to use pointer semantics instead of value semantics?  Is there a particular use case that makes us prefer them here?

src/restruct_poc/model/model.hpp
faisal-bhuiyan: My opinion is we don't want to make copies of the `Model` class objects (either shallow or deep) since it's a top-level data structure containing lots of resources. Should we wait for a use case to arise before we permit this?

faisal-bhuiyan:
> Why do we want to use pointer semantics instead of value semantics? Is there a particular use case that makes us prefer them here?

In my opinion, it is advantageous to use pointer semantics for the `Model` class for several reasons
- We could create the different attributes anywhere and don't need to copy/move them into the `Model` class, just creating a handle to them suffices
- Pointer-centric design would allow us to support Polymorphism, which for example should simplify the `Constraint`s-related mechanisms
- It will be less expensive if we decide to allow (shallow) copies of the `Model` class objects

src/restruct_poc/model/model.hpp
ddement: I'm of the opposite persuasion - unless we have good reason, we shouldn't touch the copy mechanics unless we have a specific reason to need to disable them.  Deleting them here also gets rid of things like the move constructor.  See: https://en.cppreference.com/w/cpp/language/rule_of_three

ddement:
> > Why do we want to use pointer semantics instead of value semantics? Is there a particular use case that makes us prefer them here?
> 
> In my opinion, it is advantageous to use pointer semantics for the `Model` class for several reasons
> 
> * We could create the different attributes anywhere and don't need to copy/move them into the `Model` class, just creating a handle to them suffices. This was actually the primary reasons we (@deslaughter) wanted to make this change, since we added methods to transform `Node` objects and using pointers provided a convenient way to support this feature.
> * Pointer-centric design would allow us to support Polymorphism, which for example should simplify the `Constraint`s-related mechanisms
> * It will be less expensive if we decide to allow (shallow) copies of the `Model` class objects

I guess I'd need to see an example to have an opinion on that case, but it seems to me that passing in a reference would suffice for transforming the node.  I don't see any use cases like this in the PR.

Depending on what you want to do with polymorphism, it may not be the best choice.  Polymorphism is not well supported on device, so anything that will have to run there should avoid it.  On host, it would depend on exactly what mechanisms you want to accomplish for if I'd prefer polymorphism or another, simpler technique.  

I'll defer to you guys on the path forward here.  I don't think I have enough context to make a good recommendation.

src/restruct_poc/model/model.hpp
faisal-bhuiyan: While I don't see any use case for copy/move operations on the `Model` class, let's just follow rule of zero here.

faisal-bhuiyan:
> I guess I'd need to see an example to have an opinion on that case, but it seems to me that passing in a reference would suffice for transforming the node. I don't see any use cases like this in the PR.

This looks like a general suggestion, but hopefully helpful.

https://stackoverflow.com/questions/892133/should-i-prefer-pointers-or-references-as-data-members

tests/unit_tests/restruct_poc/test_cantilever_beam.cpp
```diff
@@ -99,8 +99,18 @@ TEST(DynamicBeamTest, CantileverBeamSineLoad) {
...
+    auto nodes_vector = std::vector<Node>{};
+    for (const auto& node : model.GetNodes()) {
+        nodes_vector.push_back(*node);
+    }
+
+    auto constraints_vector = std::vector<Constraint>{};
+    for (const auto& constraint : model.GetConstraints()) {
+        constraints_vector.push_back(*constraint);
+    }
+
```
deslaughter, line 111: Could the solver constructor be changed to take what is returned by `model.GetConstraints` and `model.GetNodes` copy doesn't need to be created manually? Either that or have `model.GetConstraints` return `std::vector<Constraint>`. I think the former would be easier to understand though.

faisal-bhuiyan: I looked into implementing this and this appears to be little more involved than what I initially thought. Let's refactor the `Solver` class in a follow up work.

faisal-bhuiyan: Never mind, I actually implemented a fix for the above in this PR.

deslaughter (approved):

PR #219, opened 2024-07-20
Use Amesos2 Sparse Direct Solver

ddement:
This commit changes the sparse solver to use Amesos2.  There are a lot of small changes to ensure that types are consistent across the code, but the meat of the work is done in solver.hpp (the setup) and solve_system.hpp (the actual solve).  This commit also removes the dense system matrix, so improves scalability quite a bit.  

Note to other developers:  This change adds a dependency on Amesos2.  If you need help setting up Trilinos for your build/test environment, let me know and I can help you to get up and running.


ddement:
@deslaughter @faisal-bhuiyan 
In addition to the normal code review, make sure to take a close look at the new variation of the IEA15Rotor test that I added to make sure that I did it right.  This version mimics our "old" implementation by setting the prescribed BC on each node directly rather than on a hub node with rigid constraints.  

With the prescribed BCs, the 300 blade case runs in ~2.5s.  With the hub, the 300 blade case runs in 15s.  The hub-based formulation seems to be much harder to solve, with most of the difference being explained by the time taken for the linear solve step.

deslaughter:
> @deslaughter @faisal-bhuiyan In addition to the normal code review, make sure to take a close look at the new variation of the IEA15Rotor test that I added to make sure that I did it right. This version mimics our "old" implementation by setting the prescribed BC on each node directly rather than on a hub node with rigid constraints.
> 
> With the prescribed BCs, the 300 blade case runs in ~2.5s. With the hub, the 300 blade case runs in 15s. The hub-based formulation seems to be much harder to solve, with most of the difference being explained by the time taken for the linear solve step.

@ddement The longer solution time seems reasonable as the system is much more complicated. When the blade root node displacement is prescribed through the constraints, then all the blades are essentially independent. However, when using rigid constraints to the hub, the blades are all tied together even though the hub node displacement is prescribed. A much more interesting test would be to use the cylindrical constraint from `RotatingBeamTest.CylindricalConstraint` which truly allows the loads from all the blades feed back into the hub and affect the whole rotor rotation.

deslaughter (approved):

PR #220, opened 2024-07-22
Check Headers for Formatting in CI

ddement:
We had previously only been checking formatting in .h and .H header files in CI.  We use .hpp for our headers.  This PR enables that check and makes the changes to our format.


deslaughter (approved):
Good catch, thanks for fixing this.

PR #221, opened 2024-07-23
Reduced the amount of scratch memory used for assembling the system.

ddement:
For large systems, this was a vast overallocation (large enough for all columns instead of just the non-zero ones).


faisal-bhuiyan (approved):
Nice catch.

PR #225, opened 2024-07-25
Rewrote cmake build system to improve organization and usability

ddement:
IMPORTANT CHANGES:

- Options have been renamed to `OpenTurbine_*` instead of `OTURB`

- Tests are now built by default

- The unit test executable now lives in `tests/unit_tests`, not the top level build directory

- The built libraries now live in `src`, not the top level build directory

- To run unit tests which use shared libraries, you must copy them into the directory where you're running your unit test executable.  For example, from your build directory, run `cp src/*.dll ./` followed by `./tests/unit_tests/openturbine_unit_tests`


deslaughter (commented):
The documentation needs to be updated to reflect the changes in option naming. It also doesn't look like the compiling documentation has been updated to include the instructions for `spack`.

ddement:
> The documentation needs to be updated to reflect the changes in option naming. It also doesn't look like the compiling documentation has been updated to include the instructions for `spack`.

I updated the documentation.  In fairness, it hadn't been updated in 1.5 years (it completely missed the KokkosKernels only days), and I wouldn't be surprised if other parts are also woefully out of date.

deslaughter (approved):

PR #228, opened 2024-07-26
Refactor controllers in first step to pass Clang-Tidy

ddement:
I won't call them perfect, but they pass Clang-Tidy and are a step in the right direction.  

I have changed the ControllerIO structure to hold entries of the correct type that we want to work with (doubles, ints, etc) and to translate them into the float* buffer when passing them to/from the shared library interface.  The result is much cleaner in application code that has to work with the controller.

There is still more to do in other parts of the code before Clang-Tidy can be enabled in CI, but I figured this would be large enough to be its own PR.


src/utilities/controllers/discon_rotor_test_controller.cpp
```diff
@@ -15,28 +15,34 @@ void PITCH_CONTROLLER(
...
+    io.pitch_command_2 = 2. * M_PI * io.time / 6.;  // Full rotation every 6 seconds
+    io.pitch_command_3 = 2. * M_PI * io.time / 9.;  // Full rotation every 9 seconds
 
     // Set failure flag to zero
     *aviFAIL = 0;
 
     // Set message to success
-    strncpy(avcMSG, "success\0", io->message_array_size);
-    avcMSG[static_cast<int>(io->message_array_size) - 1] = 0;
+    strncpy(avcMSG, "success\0", io.message_array_size);
```
deslaughter, line 32: `strncpy` alone does not guarantee that the string will be null terminated; hence the following line. https://stackoverflow.com/questions/1453876/why-does-strncpy-not-null-terminate

src/utilities/controllers/turbine_controller.cpp
```diff
@@ -8,24 +8,25 @@
...
+    std::string shared_lib_path, std::string controller_function_name, std::string input_file_path,
+    std::string output_file_path
 )
-    : input_file_path_(input_file_path),
-      output_file_path_(output_file_path),
-      shared_lib_path_(shared_lib_path),
-      controller_function_name_(controller_function_name),
-      lib_(shared_lib_path, util::dylib::no_filename_decorations) {
+    : io{},
+      input_file_path_(std::move(input_file_path)),
```
deslaughter, line 15: So is is the idea with `std::move` that `input_file_path` is already a copy of the original argument which will be moved into `input_file_path_`? Is the compiler not smart enough to figure this out itself?

src/utilities/controllers/controller_io.hpp
```diff
@@ -1,100 +1,117 @@
...
+    int pitch_override = 0;                    // (55) Pitch override: 0=yes
+    int torque_override = 0;                   // (56) Torque override: 0=yes
+    size_t n_blades = 0U;                      // (61) Number of blades
+    size_t n_log_variables = 0U;               // (65) Number of variables returned for logging
+    double generator_startup_resistance = 0.;  // (72) Generator start-up resistance
+    int loads_request = 0;                     // (79) Request for loads: 0=none
+    int variable_slip_status = 0;              // (80) Variable slip current status
+    int variable_slip_demand = 0;              // (81) Variable slip current demand
+
+    void CopyToSwapArray(std::array<float, kSwapArraySize>& swap_array) const {
```
deslaughter, line 50: These copy to/from swap array functions look like they're going to be difficult to maintain with the same variable needing to be in three different locations and the correct array indices added in two places. I thought you had been talking about pushing the swap array into Device memory and having the constraints access it more directly.

src/utilities/controllers/turbine_controller.cpp
ddement: Yes, this is the idea.  And no, in most cases it is not.  In general, there's no promise that copy and move constructors don't have side effects, so the compiler is going to be limited in its ability to just pick the right one.  "Copy-and-move" has become pretty idiomatic modern C++ initialization, hence why Clang-Tidy checks for it.

src/utilities/controllers/controller_io.hpp
ddement: We'll have to add the mapping as we add new features that we support, but that should be pretty "write once, change never."  A test for each entry should be pretty easy to write, thus proving correctness.  I agree that it's verbose, but I think it's worth it to remove the need for a reinterpret cast of raw memory allocated by another codebase.  

I am still thinking about doing something along those lines, but those changes would be pretty involved, so I'm leaving it for a future step.  The goal here was to be otherwise transparent to our current usage.

src/utilities/controllers/discon_rotor_test_controller.cpp
ddement: Doesn't the `\0` take care of this already?

deslaughter: If 'avcMSG' is shorter than 'success' it will fill the buffer with characters but not add the null termination. This is one of the issues with C string functions.

ddement: Fair enough.  I've added a version of the fix back in that clang-tidy seems to like.  

If you find this version objectionable, I might just turn off the check which complains about pointer arithmetic and return to the old fix.  These things are supposed to make it easier for us to write good code, not make us jump through hoops.

deslaughter (approved):

PR #235, opened 2024-08-02
Refactored calculate_constraint_residual_gradient.hpp to pass Clang-Tidy

ddement:
This code refactors the CalculateConstraintResidualGradient kernel to pass Clang-Tidy checks.  It does this, in part, by refactoring it into one function for each type of constraint.  Each function is free from conditionals, so it should be much easier to understand what's going on for any given constraint, and it should be a little more performant on GPUs.  It also doesn't allocate stack memory for auxiliary variables that aren't used for that constraint.


faisal-bhuiyan:
Looks good -- just curious about the scope of the clang-tidy checks, Is it not acting on the whole codebase?

faisal-bhuiyan (approved):

ddement:
> Looks good -- just curious about the scope of the clang-tidy checks, Is it not acting on the whole codebase?

@faisal-bhuiyan 
Before this commit, it was only acting on .cpp files.  This commit lets it look at .hpp files, too (That's the change in the .clang-tidy file).  It was an oversight of mine when I was first setting up these checks however many months ago, but it also made a good way to divide up the work of fixing the warnings.

PR #244, opened 2024-08-15
Adds unit tests for the kernels used for sparse matrix construction

ddement:
This should provide a good base of tests for when we go to modify/extend our sparsity patterns in the future.


src/restruct_poc/solver/constraints.hpp
```diff
@@ -47,7 +48,7 @@ struct Constraints {
     View_N Phi;                                    //< Residual vector
     Kokkos::View<double* [6][12]> gradient_terms;  //< Gradient terms
 
-    Constraints(const std::vector<std::shared_ptr<Constraint>>& constraints)
+    explicit Constraints(const std::vector<std::shared_ptr<Constraint>>& constraints)
```
faisal-bhuiyan, line 51: Why is this required?

faisal-bhuiyan (approved):

src/restruct_poc/solver/constraints.hpp
ddement: You should put `explicit` ahead of constructors which take only a single entry to prevent implicit conversion.  i.e. we don't want the following code to be valid:
```
auto my_vector = std::vector<std::shared_ptr<Constraint>>>{};
Constraints my_constraints = my_vector;
 ```
One of the  sanitizers picked this up. They seem to get false negatives (missing code they should have warned about) under certain conditions, so we might find stray warnings like this for otherwise unrelated code in the future.  Just clean them up as they pop up and we should get them all taken care of in short order.

PR #250, opened 2024-08-20
Moved data structures in Beams from node-based to per-element-per-node based

ddement:
This change streamlines some of our interpolation code and makes the `Beams` data structure more self-consistent.  It will also simplify some future work, including outer loop SIMD.


src/restruct_poc/beams/beams.hpp
```diff
@@ -31,24 +31,24 @@ struct Beams {
     size_t max_elem_nodes;
     size_t max_elem_qps;
 
-    Kokkos::View<ElemIndices*> elem_indices;   // View of element node and qp indices into views
-    Kokkos::View<size_t*> node_state_indices;  // State row index for each node
+    Kokkos::View<ElemIndices*> elem_indices;    // View of element node and qp indices into views
+    Kokkos::View<size_t**> node_state_indices;  // State row index for each node
```
faisal-bhuiyan, line 35: Looks like the comments need to be updated throughout? Would be nice to add a little introduction to the `Beams` class as well.

src/restruct_poc/beams/calculate_jacobian.hpp
```diff
@@ -12,9 +12,10 @@ namespace openturbine {
 struct CalculateJacobian {
     Kokkos::View<Beams::ElemIndices*>::const_type elem_indices;  // Element indices
     Kokkos::View<double***>::const_type shape_derivative;        // Num Nodes x Num Quadrature points
-    View_Nx7::const_type node_position_rotation;        // Node global position/rotation vector
+    Kokkos::View<double** [7]>::const_type
```
faisal-bhuiyan, line 15: Just curious -- is the type here same as `Kokkos::View<const double** [7]>`?

src/restruct_poc/beams/calculate_jacobian.hpp
```diff
@@ -12,9 +12,10 @@ namespace openturbine {
...
+        node_position_rotation;                         // Node global position/rotation vector
     Kokkos::View<double** [3]> qp_position_derivative;  // quadrature point position derivative
-    View_NxN qp_jacobian;                               // Jacobians
+    Kokkos::View<double**> qp_jacobian;                 // Jacobians
```
faisal-bhuiyan, line 18: How is this change doing anything different?

faisal-bhuiyan (approved):
Looks good to me.

src/restruct_poc/beams/beams.hpp
```diff
@@ -31,24 +31,24 @@ struct Beams {
...
-    View_Nx6 node_FI;      // Inertial forces
-    View_Nx6 node_FX;      // External forces
+    Kokkos::View<double** [7]> node_x0;      // Inital position/rotation
+    Kokkos::View<double** [7]> node_u;       // State: translation/rotation displacement
+    Kokkos::View<double** [6]> node_u_dot;   // State: translation/rotation velocity
+    Kokkos::View<double** [6]> node_u_ddot;  // State: translation/rotation acceleration
+    Kokkos::View<double** [6]> node_FE;      // Elastic forces
+    Kokkos::View<double** [6]> node_FG;      // Gravity forces
+    Kokkos::View<double** [6]> node_FI;      // Inertial forces
+    Kokkos::View<double** [6]> node_FX;      // External forces
```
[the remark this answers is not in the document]
ddement: I'm not a fan of doing that for our lower level code.  

Any time we have to make changes, such as changing the dimensions of a variable, changing the layout, etc, we have to either add a new entry to types for a very specific use case or have a mix of locally defined and pre-defined.  It also makes it impossible for us to template on these parameters, which is something we may want to do in the future.  Using the full types here keeps things consistent.  

Also, our kernels do all sorts of fancy stuff with Kokkos - knowing and controlling the details locally makes sense.  In the higher level code (model, controller, etc) it makes more sense to use the aliases.

src/restruct_poc/beams/calculate_jacobian.hpp
ddement: This is a consistency change.  Everything else in the kernel uses `Kokkos::View<...>`, so it makes it easier to read and compare when going down the list of variables.

src/restruct_poc/beams/beams.hpp
ddement: Good Idea.  I'll do that

src/restruct_poc/beams/beams.hpp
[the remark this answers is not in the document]
faisal-bhuiyan: Sounds good, I must have missed the distinction between higher and lower level codes there.

PR #251, opened 2024-08-20
Removed ElemIndices data structure with two Views of size_t

ddement:
After recent changes, most of `ElemIndices` was unused.  Also, in many cases, only one piece of information was needed in a given location.  This simpler formation better communicates to the user what information is needed where.

In theory, it seems like it should be a hair faster than the old implementation, but any differences one way or the other are expected to be unmeasurable.

### DIFF
--- a/tests/unit_tests/restruct_poc/system/test_integrate_stiffness_matrix.cpp
+++ b/tests/unit_tests/restruct_poc/system/test_integrate_stiffness_matrix.cpp
@@ -18,8 +18,8 @@ void TestIntegrateStiffnessMatrix_1Element1Node1QP(
     const Kokkos::View<const double[1][6][6]>& qp_Ouu,
     const Kokkos::View<const double[1][6][6]>& qp_Quu, const std::array<double, 36>& exact_M_data
 ) {
-    constexpr auto number_of_nodes = 1;
-    constexpr auto number_of_qps = 1;
+    constexpr auto number_of_nodes = size_t{1U};
+    constexpr auto number_of_qps = size_t{1U};
 
     const auto qp_weights = get_qp_weights<number_of_qps>({2.});
     const auto qp_jacobian = get_qp_jacobian<number_of_qps>({3.});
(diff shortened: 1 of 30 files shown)


faisal-bhuiyan:
> After recent changes, most of `ElemIndices` was unused. Also, in many cases, only one piece of information was needed in a given location. This simpler formation better communicates to the user what information is needed where.

What were the pieces that were split out here? Would have been nice to get a preview.

tests/unit_tests/restruct_poc/system/test_integrate_stiffness_matrix.cpp
```diff
@@ -496,8 +452,8 @@ void TestIntegrateStiffnessMatrix_1Element1Node2QPs(
     const Kokkos::View<const double[2][6][6]>& qp_Ouu,
     const Kokkos::View<const double[2][6][6]>& qp_Quu, const std::array<double, 36>& exact_M_data
 ) {
-    constexpr auto number_of_nodes = 1;
-    constexpr auto number_of_qps = 2;
+    constexpr auto number_of_nodes = size_t{1U};
+    constexpr auto number_of_qps = size_t{2U};
```
faisal-bhuiyan, line 456: Would this be cleaner?

```
constexpr size_t number_of_nodes{1};
constexpr size_t number_of_qps{2};
```

faisal-bhuiyan (approved):
Whole lot of LOC but mostly resulting from the interface change to `Beams`. Overall, it looks good to me.

tests/unit_tests/restruct_poc/system/test_integrate_stiffness_matrix.cpp
ddement: Depends on who you ask.  I like using auto and making type defined by what appears on the right-hand side of the expression.  This helps keywords and variable names line up better for complex types.  It also makes the form of the code uniform regardless of your initialization strategy (copy assignment vs constructor vs factor function all look the same: `auto var_name = make_my_var()`).  I admit it looks a little chunky for really simple uses like this, but I like that it provides one uniform and organized way to declare a variable in all cases and all variable types.  

I think C++20 or 23 gives us a literal for `size_t` like the U for `unsigned`.  Once we can use that, it will make this type of code look better.

ddement:
> > After recent changes, most of `ElemIndices` was unused. Also, in many cases, only one piece of information was needed in a given location. This simpler formation better communicates to the user what information is needed where.
> 
> What were the pieces that were split out here? Would have been nice to get a preview.

@faisal-bhuiyan 

My wording should have been better here.  What I meant was that instead of having a single View of a struct containing `num_nodes` and `num_qps`, I made it two Views: `num_nodes_per_elem` and `num_qps_per_elem`.  These are still owned by `Beams`.

deslaughter (commented):
Enable VTK output and update code affected by these changes.

faisal-bhuiyan:
> Enable VTK output and update code affected by these changes.

I think we should address #256 as a priority follow-up item so this does not keep happening.

ddement:
> > Enable VTK output and update code affected by these changes.
> 
> I think we should address #256 as a priority follow-up item so this does not keep happening.

@faisal-bhuiyan @deslaughter 
Agreed.  I will add VTK to CI and update the output to work with these changes as soon as Faisal's PR #254 and this commit are both merged in.  Faisal's PR does a lot of what is necessary already, so I don't want to duplicate that work just to create merge conflicts.

PR #253, opened 2024-08-20
Add generator torque to the azimuth DOF

faisal-bhuiyan:
Adds logic to apply generator torque to the turbine via a revolute joint. Following tasks were accomplished here

- `Constraint` class was refactored to create modularity in its c-tor
- Interface for the `AddRevoluteJointConstraint` method in `Model` class was refactored to accept an axis and torque input as a precursor for applying the generator torque
- `AssembleConstraints` kernel was enhanced to add the constraint forces applied to a revolute joint during assembly of the residual vector
- New kernel `AssembleConstraintForce` -> `CalculateRevoluteJointForce` developed to calculate the modifications required in the residual vector to accommodate the applied torque via the constraint
- VTK method `BeamsWriteVTK` was updated to work with the changes to solver interface
- New unit test (`RotatingBeamTest, GeneratorTorque`) added for all the new logic


src/restruct_poc/solver/constraints.hpp
```diff
@@ -29,19 +26,21 @@ struct Constraints {
         double axis_x[3];          // Unit vector representing x rotation axis
         double axis_y[3];          // Unit vector representing y rotation axis
         double axis_z[3];          // Unit vector representing z rotation axis
+        double* control;           //< Control signal
```
ddement, line 29: Delete this variable.  It can't work on GPU (a pointer to host memory can't be used) and is only used as a flag (it doesn't look like it's ever set)

src/restruct_poc/solver/assemble_constraint_force.hpp
```diff
@@ -0,0 +1,58 @@
...
+    View_N::const_type control;
+    View_Nx7::const_type node_u;
+    View_N R_system;
+
+    KOKKOS_FUNCTION
+    void operator()(const int i_constraint) const {
+        const auto& cd = data(i_constraint);
+
+        // Applies the torque from a revolute joint constraint to the system residual
+        if (cd.type == ConstraintType::kRevoluteJoint && cd.control) {
```
ddement, line 24: There's no need to check cd.control here (plus, it should just be deleted from the struct, as I recommended in another comment).

src/restruct_poc/solver/assemble_constraint_force.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+#include "constraints.hpp"
+
+#include "src/restruct_poc/math/quaternion_operations.hpp"
+#include "src/restruct_poc/math/vector_operations.hpp"
+#include "src/restruct_poc/types.hpp"
+
+namespace openturbine {
+
+struct AssembleConstraintForce {
```
ddement, line 13: I'm thinking we should go ahead and split this function in a similar way to how I did calculate_constraint_residual_gradient (i.e. turn the contents of the if condition into its own kernel).  I know it might seem silly with just one constraint type, but it will set us up for adding more types in the future and makes testing nice.

tests/unit_tests/restruct_poc/test_rotating_beam.cpp
```diff
@@ -628,4 +632,104 @@ TEST(RotatingBeamTest, CylindricalConstraint) {
     );
 }
 
+TEST(RotatingBeamTest, GeneratorTorque) {
```
ddement, line 635: Similar to the test at the top of this file, it would be good to add a check on the state after the 10 steps are complete to ensure that the results aren't changing as we change the code in the future.

ddement (commented):
It would be good to add a unit test for the new assemble_constraint_force kernel. 

Similarly, do we have any data (an analytical result or other code comparison) that we can compare to verify that we're getting correct results?

ddement (approved):
Looks good to me, assuming you've validated the results.

deslaughter:
When I set `tilt = 0.` in the `GeneratorTorque` test, the beam doesn't move. Can you reproduce this?

faisal-bhuiyan:
> When I set `tilt = 0.` in the `GeneratorTorque` test, the beam doesn't move. Can you reproduce this?

I believe we fixed this problem and put in place couple of unit tests that examines the rotational behavior of the rotor with the applied torque. Does that look good?

deslaughter (approved):

PR #254, opened 2024-08-28
Improved consistency of system assembly code

ddement:
This PR consists of several commits to clean up how our system assembly is performed to improve code organization and data flow.  The biggest practical changes are

- state and constraints are no longer owned by the solver class
- the constraints class no longer uses HostData and DeviceData structures
- the solver folder was also broken up into the solver, constraints, state, and step folders, since solver had kind of become a grab bag of these four different concepts

The easiest way to review it is to look at each commit individually - they're all pretty self-contained ideas.


src/restruct_poc/constraints/calculate_constraint_residual_gradient.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+namespace openturbine {
+
+struct CalculateConstraintResidualGradient {
+    Kokkos::View<ConstraintType*>::const_type type;
+    Kokkos::View<size_t* [2]>::const_type node_index;
+    Kokkos::View<size_t* [2]>::const_type row_range;
+    Kokkos::View<size_t* [2][2]>::const_type node_col_range;
+    Kokkos::View<double* [3]>::const_type X0_;
+    Kokkos::View<double* [3][3]>::const_type axis;
```
deslaughter, line 24: @ddement changing `x_axis`, `y_axis`, and `z_axis` to a 3x3 matrix without comments makes it very hard to know what data is in there. The name `axis` implies a 3 element vector. This change significantly reduces the readability of this code and makes it harder to understand what is happening. And yes, I see that the matrix is unpacked into vectors before use, but it's just another layer of obscurity when somebody is passing a view called `axis` into `CalculateConstraintResidualGradient`. At the very least this should be renamed to `axes` since it is a view of multiple axes.

src/restruct_poc/constraints/calculate_constraint_residual_gradient.hpp
```diff
@@ -0,0 +1,57 @@
...
+    View_N::const_type control;
+    View_Nx7::const_type constraint_u;
+    View_Nx7::const_type node_u;
+    View_N Phi_;
+    Kokkos::View<double* [6][12]> gradient_terms;
+
+    KOKKOS_FUNCTION
+    void operator()(const int i_constraint) const {
+        if (type(i_constraint) == ConstraintType::kFixedBC) {
```
deslaughter, line 33: Is there a reason this isn't a `switch` statement?

src/restruct_poc/constraints/calculate_cylindrical_constraint.hpp
```diff
@@ -101,8 +110,12 @@ struct CalculateCylindricalConstraint {
         //---------------------------------
         {
             // Extract gradient block for target node of this constraint
-            const auto B =
-                Kokkos::subview(gradient_terms, i_constraint, Kokkos::ALL, cd.target_node_col_range);
+            const auto B = Kokkos::subview(
+                gradient_terms, i_constraint, Kokkos::ALL,
+                Kokkos::make_pair(
+                    node_col_range(i_constraint, 1, 0), node_col_range(i_constraint, 1, 1)
```
deslaughter, line 116: Again, this change significantly reduces readability of the code. Where is it documented that `node_col_range(i_constraint, 1, 0)` refers to the target node and `node_col_range(i_constraint, 0, 0)` refers to the base node? It's not clear from scanning through this code. I don't see how calling `make_pair` every time this is run is more efficient than calculating it up front and giving it a descriptive name.

src/restruct_poc/constraints/calculate_prescribed_bc_constraint.hpp
```diff
@@ -76,7 +79,9 @@ struct CalculatePrescribedBCConstraint {
         //----------------------------------------------------------------------
 
         // Extract residual rows relevant to this constraint
-        const auto Phi = Kokkos::subview(Phi_, cd.row_range);
+        const auto Phi = Kokkos::subview(
```
deslaughter, line 82: This same code is repeated through in 6 files. If it ever needs to be modified then it will need to be updated 6 times. A new developer is almost guaranteed to miss one of the spots where it needs to be updated. This should be moved to the calling function and passed in since all of the calculate constraint functions need it.

src/restruct_poc/step/update_system_variables.hpp
```diff
@@ -3,19 +3,21 @@
...
+#include "src/restruct_poc/system/calculate_quadrature_point_values.hpp"
+#include "src/restruct_poc/system/calculate_tangent_operator.hpp"
+#include "src/restruct_poc/system/update_node_state.hpp"
 
 namespace openturbine {
 
-inline void UpdateState(
-    const Beams& beams, const View_Nx7& Q, const View_Nx6& V, const View_Nx6& A
-) {
+inline void UpdateSystemVariables(StepParameters& parameters, const Beams& beams, State& state) {
```
deslaughter, line 20: This function is doing a lot more than updating system variables as it's name implies. I wouldn't expect it to be doing any calculations. What is considered a system variable?

src/restruct_poc/constraints/calculate_constraint_residual_gradient.hpp
ddement: Style.  It's not obvious which one will be faster (or that the speed difference will be significant).  I tend to prefer `if` statements, since they have no chance for fall-through behavior (thus are more readable and less bug prone) and compilers/hardware tend to be well optimized for them.

src/restruct_poc/constraints/calculate_constraint_residual_gradient.hpp
ddement: I agree renaming it will make things clearer, I'll do that.  

I packed them into a single 3x3 entry because we typically want all three of them when we want one of them.  This not only simplifies calling kernels that use them, but it also improves caching/coalescing behavior.

src/restruct_poc/constraints/calculate_cylindrical_constraint.hpp
ddement: I'll look into cleaning this up.  I'll split it into `base_node_col_range` and `target_node_col_range` and use pairs in them.

I chose to go to local `make_pair` because it is very cheap (just copying 2 `size_t` from heap memory to stack memory), and this makes the load properly coalesced on GPU (the entries are further away from each other).

src/restruct_poc/constraints/calculate_prescribed_bc_constraint.hpp
ddement: Similar to above, I'll go back to the row_range being a pair, so the code will look much the same as it did before.

Are you suggesting that I write a function like this?
```
auto GetLocalPhi(Kokkos::View<double*> Phi_, Kokkos::pair<size_t, size_t> row_range) {
    return Kokkos::subview(Phi_, row_range);
}
```

src/restruct_poc/step/update_system_variables.hpp
ddement: A system variable is one that is owned by the beams class.  We have three types of variables currently: system, state, and constraints.  System variables are per-element and reside in the beams class.  State variables are per-node and reside in the state class.  Constraint variables are per-constraint and reside in the constraints class.  I didn't rename beams to be system, but we could do that if you think it would help.

In truth, the function is all about computing each element's contributions to the residual, stiffness, and inertia matrices.    The others, for now at least, are just working variables.  

It also computes each node's tangent matrix, which lives in `state`.  This is inconsistent, and I can split that out into its own function.  I had several practical reasons for doing it this way, but in hindsight they are probably not good enough.  

If you look in the step function where this is called, the name `UpdateSystemVariables` provides a connection to `AssembleSystem*` and symmetry with the constraint operations (`UpdateConstraintVariables` and `AssembleConstraints*`) and then state operations.  The `Update*Variables` functions compute the individual contributions to the system, while the `Assemble*` functions assemble them into the linear system.  

I know it isn't perfect, but the idea is to make a couple steps in the right direction.  I'm open to discussion if you have some ideas.

src/restruct_poc/constraints/calculate_constraint_residual_gradient.hpp
faisal-bhuiyan: Not sure I agree with the readability comment -- I find `switch` statements to be cleaner and more readable than an equivalent nested `if-else if-else` block. The cpp-coreguidelines appear to have the same outlook.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-switch-if

ddement: My comment on fall-through is about the need for a `break` at the end of each case.  For example
```
switch(type) {
  case type_1 :
    std::cout << 1;
  case type_2:
    std::cout << 2;
    break;
  case type_3:
    std::cout << 3;
    break;
}
```

If my `type` is `type_1`, then it will print 12, but if my `type` is `type_2` it will just print 2.  I know that you can write this type of code intentionally, but I would contend that it is harder to reason about - not only do I have to remember that the fall through happens, but I have to determine if it was intentional.  Compare this to:
```
if(type == type_1) {
  std::cout << 12;
}
if(type == type_2) {
  std::cout << 2;
}
if(type == type_3) {
  std::cout << 3;
}
```
Here, I know exactly what is happening for each case and there is no ambiguity of if the behavior for `type_1` is a bug.  

Looking at the page you linked, it does mention this in ES.78.  I agree with "always use `break` or `[[fallthrough]`. 
 However, I disagree with their suggestion of using multiple labels for a single statement.  Coming back later, it is hard to determine if you really meant for fall through or wanted to do nothing for that case and forgot a `break`.
 
 However, if both of you agree that we should you switch statements in this case, I will make the change.

deslaughter: You can leave it as `if` statements. I'm more used to languages that don't use the implicit fallthrough in their switches like Fortran and Go. It is an easy construct to mess up in C/C++.

src/restruct_poc/constraints/calculate_cylindrical_constraint.hpp
deslaughter: Yes, using `base_node_col_range` and `target_node_col_range` makes it much easier to read the code and understand the intent.

src/restruct_poc/constraints/calculate_prescribed_bc_constraint.hpp
deslaughter: I was thinking the function with the `if` statements for each constraint type could go ahead and calculate the subview of `Phi` and pass it in to the constraint calculation function. Same could be done with the base and target node blocks. It keeps the code from being repeated and should keep the constraint functions cleaner.

src/restruct_poc/step/update_system_variables.hpp
deslaughter: Thanks for the explanation on the system variables. I'm not sure that I agree with the logic on calling them variables, but I'll need to try working with it before I'll know if it should change.

deslaughter (approved):

src/restruct_poc/constraints/calculate_prescribed_bc_constraint.hpp
ddement: Alternative idea to achieve the same thing: make Phi into a View with dimensions NumConstraints x 6, similar to the gradient_terms matrix.  We could then just assign into Phi by doing `Phi_(i_constraint, i) = values` without the need for subviews.  The logic of copying to the correct spot in our matrix using row_range would then be moved to `AssembleConstraintsResidual`, just like for the system residual.  We could do something similar by splitting `gradient_terms` into `target_gradient_terms` and `base_gradient_terms`.

These changes would remove our dependence on the ranges in the physics kernels (a good thing) and move them to our Assembly where they belong.  I want to avoid passing subviews of multidimensional views around, when possible, since we then have to either template on their type or use the less efficient (and more verbose) `LayoutStride`.

src/restruct_poc/constraints/calculate_constraint_residual_gradient.hpp
faisal-bhuiyan: Not a huge fan of the fall-through behavior myself either and tend to use this with caution and always with `break` everywhere else to make intentions clear. I personally think `switch` is neater than clunky-looking conditional branching in many instances.

src/restruct_poc/constraints/calculate_prescribed_bc_constraint.hpp
deslaughter: That sounds like a good solution. The constraints residual won't ever have more than 6 items in it and that keeps it consistent with the gradient blocks.

PR #257, opened 2024-09-04
Changed the formatting CI workflow to no longer need the DoozyX action

ddement:
The CI action we used to check formatting broke recently.  This change should make our CI process more robust to outside changes, since it only uses the "checkout" action

Note that we are now using clang-format version 15 (vs 16 before).  This is the latest version available on the ubuntu-latest runners.  We could jump through more hoops to install a newer version, but 15 should suffice.

### DIFF
--- a/.github/workflows/formatting.yaml
+++ b/.github/workflows/formatting.yaml
@@ -3,15 +3,24 @@ name: OpenTurbine-CI
 on: push
 
 jobs:
-  ClangFormat:
+  clang-format:
     runs-on: ubuntu-latest
+
     steps:
-    - name: Clone
+    - name: Checkout repository
       uses: actions/checkout@v4
-    - name: Check formatting
-      uses: DoozyX/clang-format-lint-action@v0.16.2
-      with:
-        source: './src ./tests/unit_tests'
-        exclude: '.'
-        extensions: 'cpp,hpp'
-        clangFormatVersion: 16
+
+    - name: Run clang-format
+      run: |
+        find . -regex '.*\.\(cpp\|hpp\|c\|h\)' -exec clang-format -i {} \;
+        
+    - name: Check for formatting changes
+      run: |
+        if [[ `git status --porcelain` ]]; then
+          echo "Code is not formatted. Please run clang-format."
+          git diff
+          exit 1
+        else
+          echo "Code is properly formatted."
+        fi
+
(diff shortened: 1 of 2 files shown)


.github/workflows/formatting.yaml
```diff
@@ -3,15 +3,24 @@ name: OpenTurbine-CI
...
-      uses: DoozyX/clang-format-lint-action@v0.16.2
-      with:
-        source: './src ./tests/unit_tests'
-        exclude: '.'
-        extensions: 'cpp,hpp'
-        clangFormatVersion: 16
+
+    - name: Run clang-format
+      run: |
+        find . -regex '.*\.\(cpp\|hpp\|c\|h\)' -exec clang-format -i {} \;
```
faisal-bhuiyan, line 15: Would the following be a little cleaner?

```
run: |
        FILES=$(find . -name '*.cpp' -o -name '*.hpp' -o -name '*.c' -o -name '*.h')
        clang-format --dry-run --Werror $FILES
```

.github/workflows/formatting.yaml
```diff
@@ -3,15 +3,24 @@ name: OpenTurbine-CI
...
+        
+    - name: Check for formatting changes
+      run: |
+        if [[ `git status --porcelain` ]]; then
+          echo "Code is not formatted. Please run clang-format."
```
faisal-bhuiyan, line 20: Would the following message make things a little clearer?

`echo "::error::Code is not properly formatted. Please run clang-format locally and commit the changes."`

.github/workflows/formatting.yaml
```diff
@@ -3,15 +3,24 @@ name: OpenTurbine-CI
 on: push
 
 jobs:
-  ClangFormat:
+  clang-format:
     runs-on: ubuntu-latest
+
```
faisal-bhuiyan, line 8: These empty lines are not used in rest of the CI workflows, just pointing it out.

faisal-bhuiyan (approved):
Looks good

.github/workflows/formatting.yaml
ddement: Maybe.  Honestly, I don't think anybody will read this message - if there's a formatting error, you'll get a big red X and the dozens of lines of git-diff.

.github/workflows/formatting.yaml
ddement: Possibly.  I'd have to try it out to make sure it behaves the same way.  I'll put it on my to-do list to try something like this out.

PR #265, opened 2024-09-05
Add openfast/aerodyn-inflow as an optional external dependency

faisal-bhuiyan:
TBD

### DIFF
--- a/Dependencies.cmake
+++ b/Dependencies.cmake
@@ -4,6 +4,34 @@ function(openturbine_setup_dependencies)
   find_package(Amesos2 REQUIRED)
   find_package(yaml-cpp REQUIRED)
 
+  # Add external project to build the ADI library from OpenFAST
+  include(ExternalProject)
+
+  # print CMKE_BINARY_DIR
+  message(STATUS "CMAKE_BINARY_DIR: ${CMAKE_BINARY_DIR}")
+
+  ExternalProject_Add(OpenFAST_ADI
+    PREFIX ${CMAKE_BINARY_DIR}/../OpenFAST_ADI
+    GIT_REPOSITORY https://github.com/OpenFAST/openfast.git
+    GIT_TAG dev
+    GIT_SHALLOW TRUE
+    GIT_SUBMODULES_RECURSE FALSE
+    CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
+               -DBUILD_SHARED_LIBS=ON                # Build shared libraries
+               -DBUILD_TESTING=OFF                   # Disable tests
+    BUILD_COMMAND ${CMAKE_COMMAND} --build . --target aerodyn_inflow_c_binding
+    INSTALL_COMMAND ""
+    BUILD_IN_SOURCE FALSE
+  )
+
+  ExternalProject_Get_Property(OpenFAST_ADI SOURCE_DIR)
+  message(STATUS "SOURCE_DIR: ${SOURCE_DIR}")
+  # Define the OpenFAST_ADI shared library for the aerodyn_inflow_c_binding
+  set_target_properties(OpenFAST_ADI PROPERTIES
+    IMPORTED_LOCATION "${SOURCE_DIR}/build/modules/aerodyn/aerodyn_inflow_c_binding/libaerodyn_inflow_c_binding.so"
+  )
+
+
   # Optionally find and link MKL if available
   if(TARGET Kokkos::MKL)
     find_package(MKL REQUIRED)


cmake/Dependencies.cmake
```diff
@@ -0,0 +1,51 @@
...
+      BUILD_IN_SOURCE OFF            # Build in a separate directory for cleaner output
+      BINARY_DIR ${CMAKE_BINARY_DIR}/OpenFAST_ADI_build
+      # Build only the aerodyn_inflow_c_binding target and do it sequentially (avoid parallel build)
+      BUILD_COMMAND ${CMAKE_COMMAND} --build . --target aerodyn_inflow_c_binding -- -j 1
+      INSTALL_COMMAND
+        ${CMAKE_COMMAND} -E copy
+        ${CMAKE_BINARY_DIR}/OpenFAST_ADI_build/modules/aerodyn/libaerodyn_inflow_c_binding${CMAKE_SHARED_LIBRARY_SUFFIX}
+        ${CMAKE_BINARY_DIR}         # Copy the library to the binary directory
+    )
+  endif()
```
ddement, line 50: How do we feel about building this as part of our CMake process vs the user bringing their own?  How would one link in their own version of OpenFAST?

tests/unit_tests/regression/test_aerodyn_inflow.cpp
```diff
@@ -0,0 +1,40 @@
+#include <gtest/gtest.h>
+
+#include "test_utilities.hpp"
+
+#include "src/vendor/dylib/dylib.hpp"
+
+namespace openturbine::tests {
+
+TEST(AerodynInflowTest, ADI_C_PreInit) {
```
ddement, line 9: For this test to run in CI, we'll need to make sure that the `libaerodyn_inflow_c_binding.*` file is in the same directory as the unit test executable.

tests/unit_tests/regression/test_aerodyn_inflow.cpp
```diff
@@ -0,0 +1,40 @@
...
+    ADI_C_PreInit(
+        &numTurbines, &transposeDCM, &debuglevel, &error_status_c,
+        static_cast<char*>(error_message_c)
+    );
+
+    EXPECT_EQ(numTurbines, 1);
+    EXPECT_EQ(transposeDCM, 1);
+    EXPECT_EQ(debuglevel, 0);
+    EXPECT_EQ(error_status_c, 0);
+    EXPECT_STREQ(static_cast<char*>(error_message_c), "");
```
ddement, line 37: Is there anything changing here that we can test?

tests/unit_tests/regression/test_aerodyn_inflow.cpp
ddement: Also, this test should be disabled if we aren't building with aerodyn support

tests/unit_tests/regression/test_aerodyn_inflow.cpp
```diff
@@ -0,0 +1,40 @@
...
+    auto path = std::string{"libaerodyn_inflow_c_binding."};
+#ifdef __APPLE__
+    path += "dylib";
+#elif __linux__
+    path += "so";
+#else if _WIN32 || _WIN64
+    path += "dll";
+#endif
+    const util::dylib lib(path, util::dylib::no_filename_decorations);
+    auto ADI_C_PreInit = lib.get_function<void(int*, int*, int*, int*, char*)>("ADI_C_PreInit");
```
ddement, line 20: If our coupling is through a shared library, is this not just testing that a shared library file exists with that function in it?  If so, we could just make our own for simple testing purposes and then require the full aerodyn one only when we actually need its functionality.

deslaughter (approved):

tests/unit_tests/regression/test_aerodyn_inflow.cpp
faisal-bhuiyan: We are testing the simplest method from the external library here, this will grow in the next PR(s) as we create interface in openturbine for calling the methods we need from aerodyn inflow library.

tests/unit_tests/regression/test_aerodyn_inflow.cpp
faisal-bhuiyan: This is calling a rather a simple method that should not change anything if we're using the interface correctly.

tests/unit_tests/regression/test_aerodyn_inflow.cpp
faisal-bhuiyan: I added conditional support for the test, will wait to see the CI outputs before making any changes to the workflow there.

cmake/Dependencies.cmake
faisal-bhuiyan: We should build this as part of openturbine rather than making any assumptions on what version user may have -- the ADI interface is subject to change as I understand from communication with the OFAST team.

ddement: That makes sense.

tests/unit_tests/regression/test_aerodyn_inflow.cpp
ddement: Since we're building aerodyn ourselves and disabling the unit tests when we don't, this should be fine

tests/unit_tests/regression/test_aerodyn_inflow.cpp
ddement: Sounds good

ddement (approved):

tests/unit_tests/regression/test_controller.cpp
```diff
@@ -16,7 +16,9 @@ TEST(ControllerTest, DisconController) {
     // at time = 0.0s
 
     // Use dylib to load the dynamic library and get access to the controller functions
-    const util::dylib lib("./DISCON.dll", util::dylib::no_filename_decorations);
+    const std::filesystem::path project_root = FindProjectRoot();
+    const std::filesystem::path full_path = project_root / "build/tests/unit_tests/DISCON.dll";
```
ddement, line 20: This won't work if your build directory isn't named build.  For example, I always have at least two: build-cpu and build-gpu and make others for different configurations when needed.  It also won't work if the build directory is not located in the project root.

tests/unit_tests/regression/test_utilities.cpp
```diff
@@ -6,6 +6,19 @@
...
+
+    while (!currentPath.empty()) {
+        if (std::filesystem::exists(currentPath / "CMakeLists.txt")) {
+            return currentPath;
+        }
+        currentPath = currentPath.parent_path();
+    }
+
+    throw std::runtime_error("Could not find project root directory. CMakeLists.txt not found.");
+}
```
ddement, line 20: Similar to my previous comment, this one will not work if the build directory is not a sub-directory of the source.

cmake/SharedLibraries.cmake
```diff
@@ -0,0 +1,13 @@
...
+  # Copy the shared library to the unit tests directory after building
+  add_custom_command(TARGET ${target_name} POST_BUILD
+    COMMAND ${CMAKE_COMMAND} -E copy
+    $<TARGET_FILE:${target_name}> ${CMAKE_BINARY_DIR}/tests/unit_tests/$<TARGET_FILE_NAME:${target_name}>
+  )
+endfunction()
+
+# Define the DISCON shared libraries for the turbine controller
+add_shared_library(DISCON)
+add_shared_library(DISCON_ROTOR_TEST_CONTROLLER)
```
ddement, line 13: Put these calls out in the src/CMakeLists.txt to keep it consistent with the way/location that other libraries/executables are defined.

cmake/SharedLibraries.cmake
```diff
@@ -0,0 +1,13 @@
+function(add_shared_library target_name)
+  add_library(${target_name} SHARED)
+  set_target_properties(${target_name} PROPERTIES PREFIX "" SUFFIX ".dll")
+  # Copy the shared library to the unit tests directory after building
+  add_custom_command(TARGET ${target_name} POST_BUILD
+    COMMAND ${CMAKE_COMMAND} -E copy
+    $<TARGET_FILE:${target_name}> ${CMAKE_BINARY_DIR}/tests/unit_tests/$<TARGET_FILE_NAME:${target_name}>
```
ddement, line 7: Since we're doing this here, delete the corresponding copy commands in the CI files

deslaughter (approved):

tests/unit_tests/regression/test_utilities.cpp
faisal-bhuiyan: Let's create an issue to harden this code later on.

tests/unit_tests/regression/test_controller.cpp
faisal-bhuiyan: I removed usage of `FindProjectRoot()` in the test files by adding a copy of the shared library files in the binary directory for macOS. This is hacky but let's take care of this edge case in a follow-up work and keep making progress to priority items.

ddement: Sounds good.

tests/unit_tests/regression/test_utilities.cpp
ddement: Agreed.

ddement (approved):

PR #272, opened 2024-09-16
Calculate current node and quadrature point position

deslaughter:
This PR adds several kernels to calculate the current node and quadrature point locations, which is needed for the following reasons
- VTK outputs
- For any future output data writing
- Interfacing with the `AeroDyn-Inflow` external library for calculating aerodynamic loads because the positions of the blade node/quadrature points will need to be passed on

This cleans up a lot of manual calculations that were being performed inside `vtkout.hpp` and instead performs them inside the calculation kernels.


src/beams/calculate_QP_deformation.hpp
```diff
@@ -0,0 +1,84 @@
...
+namespace openturbine {
+
+struct CalculateQPDeformation {
+    Kokkos::View<size_t*>::const_type qp_num_qps_per_element;
+    Kokkos::View<double** [3]>::const_type qp_x0_;
+    Kokkos::View<double** [4]>::const_type qp_r_;
+    Kokkos::View<double** [7]>::const_type qp_x_;
+    Kokkos::View<double** [3]> qp_deformation_;
+
+    KOKKOS_FUNCTION void operator()(const size_t i_elem, const size_t i_qp) const {
```
ddement, line 17: I've been doing these types of loops using `TeamPolicy` and hierarchical parallelism.  It gets similar or better performance, lets us write unit tests based on just the per-element action, and avoids the need for the early exit condition.

This doesn't seem to be a performance critical loop, so up to you if you want to change it.

src/beams/create_beams.hpp
```diff
@@ -119,8 +119,8 @@ inline Beams CreateBeams(const BeamsInput& beams_input) {
         InterpolateToQuadraturePoints{
             beams.num_nodes_per_element, beams.num_qps_per_element, beams.shape_interp,
             beams.shape_deriv, beams.qp_jacobian, beams.node_u, beams.node_u_dot, beams.node_u_ddot,
-            beams.qp_u, beams.qp_u_prime, beams.qp_r, beams.qp_r_prime, beams.qp_u_dot,
-            beams.qp_omega, beams.qp_u_ddot, beams.qp_omega_dot}
+            beams.qp_x0, beams.qp_r0, beams.qp_u, beams.qp_u_prime, beams.qp_r, beams.qp_r_prime,
+            beams.qp_u_dot, beams.qp_omega, beams.qp_u_ddot, beams.qp_omega_dot, beams.qp_x}
```
ddement, line 123: Not related to this PR explicitly: do we want to have more readable names for our variables in beams/state?  Short aliases in our kernels are nice for reading the math, but having them here can make it harder for me to keep track of everything.  I defer that decision (and what the names should be) to you and Faisal.

src/model/model.hpp
```diff
@@ -144,6 +146,13 @@ class Model {
     /// Returns the number of constraints in the model
     [[nodiscard]] size_t NumConstraints() const { return this->constraints_.size(); }
 
+    // Returns a State object initialized from the model nodes
+    State CreateState() {
```
ddement, line 150: This should be `[[nodiscard]]`

src/system/calculate_quadrature_point_values.hpp
```diff
@@ -58,14 +61,15 @@ struct CalculateQuadraturePointValues {
...
-        );
         Kokkos::parallel_for(
             Kokkos::TeamThreadRange(member, num_qps),
             CalculateTemporaryVariables{i_elem, qp_x0_prime_, qp_u_prime_, qp_x0pupSS_}
         );
         member.team_barrier();
+        Kokkos::parallel_for(
+            Kokkos::TeamThreadRange(member, num_qps), CalculateRR0{i_elem, qp_x_, qp_RR0_}
+        );
+        member.team_barrier();
```
ddement, line 72: Why did you change the order of `CalculateTemporaryVariables` and `CalculateRR0` and add the new team_barrier?  I'm not seeing the change in dependencies.

tests/unit_tests/regression/test_rotor.cpp
```diff
@@ -49,7 +48,7 @@ void WriteMatrixToFile(const std::vector<std::vector<T>>& data, const std::strin
 
 TEST(RotorTest, IEA15Rotor) {
     // Flag to write output
-    constexpr bool write_output(false);
+    constexpr bool write_output(true);
```
ddement, line 51: did you mean to push this change to main?

tests/unit_tests/regression/test_rotor.cpp
```diff
@@ -62,7 +61,7 @@ TEST(RotorTest, IEA15Rotor) {
     constexpr size_t max_iter(6);
     constexpr double step_size(0.01);  // seconds
     constexpr double rho_inf(0.0);
-    constexpr double t_end(0.1);
+    constexpr double t_end(10);
```
ddement, line 64: Same as above - looks like a change just for development.

tests/unit_tests/system/test_calculate_RR0.cpp
```diff
@@ -39,34 +34,31 @@ TEST(CalculateRR0Tests, OneNode) {
...
-                   0.,      0.,     0.,      2780.,  4400.,   -400.,   0.,      0.,      0.,
-                   -3280.,  2372.,  3296.,   0.,     0.,      0.,      2960.,   -1504.,  4028.,
-
-                   16412.,  74896., -11984., 0.,     0.,      0.,      -27376., 17284.,  70528.,
-                   0.,      0.,     0.,      70736., -10688., 30076.,  0.,      0.,      0.,
-                   0.,      0.,     0.,      16412., 74896.,  -11984., 0.,      0.,      0.,
-                   -27376., 17284., 70528.,  0.,     0.,      0.,      70736.,  -10688., 30076.};
+    Kokkos::parallel_for("CalculateRR0", 2, CalculateRR0{0, x, rr0});
+
+    constexpr auto expected_rr0_data = std::array{-44.,  4.,   118., 0.,    0.,   0.,    //
```
ddement, line 49: Does the `//` prevent the clang-format from ruining the matrix layouts?  If so, that is magical.

ddement (approved):

src/beams/calculate_QP_deformation.hpp
deslaughter: I think it makes sense to do the hierarchical parallelism, but I didn't immediately know how to do it in this context and I needed to make sure this was working. Since this is currently only called from `vtkouput` and not in the main loop I'll add a note that this could be improved later. Or you're welcome to add a commit for it.

src/model/model.hpp
deslaughter: I don't disagree, could you explain what it means in this context? I thought this tag was so a function that isn't called wasn't discarded.

src/system/calculate_quadrature_point_values.hpp
deslaughter: That's a good catch. I had made the change in the first commit and the second commit made the change unnecessary. I had added the function to calculate QP position, `qp_x` right before this, which made `CalculateRR0` depend on that function's completion. So I added the team barrier to guarantee that the positions were all calculated before moving on. Then I ended up moving the calculate QP position call which made this change unnecessary. I'll fix it in the next commit.

tests/unit_tests/regression/test_rotor.cpp
deslaughter: You're right, I didn't mean to push this.

tests/unit_tests/regression/test_rotor.cpp
deslaughter: Or this

tests/unit_tests/system/test_calculate_RR0.cpp
deslaughter: Yes, the `//` forces clang-format to keep the line break where it should be, which makes the matrix so much more readable.

src/model/model.hpp
ddement: nodiscard means that it is a compile time error to call this function and not store the returned value.  So `model.CreateState();` is an error, but `auto state = model.CreateState();` is not.  Since this function doesn't have side effects, it can be assumed that discarding the return value is a mistake by the user.  

I think this method can actually have a trailing const, too, since it doesn't modify model.

src/beams/create_beams.hpp
ddement: Sounds good.  I agree for the parts that actually do the math - staying closer to the theory makes it much easier to follow/read.  For these higher-level ones, I go back and forth between preferring more human readable names and sticking to theory so it's easy to visually trace them through the code.

src/beams/calculate_QP_deformation.hpp
ddement: I'll make a note to circle back to this one in a follow up PR - I have a bit of state built up in my current workspace and don't want to hold you up.

src/model/model.hpp
deslaughter: Thanks for the explanation, I understand why you'd want the `[[nodiscard]]` here. I've added the `const` postfix.

src/model/model.hpp
```diff
@@ -144,6 +146,13 @@ class Model {
     /// Returns the number of constraints in the model
     [[nodiscard]] size_t NumConstraints() const { return this->constraints_.size(); }
 
+    // Returns a State object initialized from the model nodes
+    [[nodiscard]] State CreateState() const {
```
faisal-bhuiyan, line 150: Just making sure -- is this returning a copy of the created `State` object or moving the object itself?

tests/unit_tests/model/test_model.cpp
```diff
@@ -168,4 +168,31 @@ TEST(Model, ModelConstructorWithPointers) {
     ASSERT_EQ(model.NumConstraints(), 0);
 }
 
+TEST(Model, ModelCreateState) {
+    Model model;
+
+    // Rotation of 1 radian around x
+    auto R1 = RotationVectorToQuaternion({1.0, 0., 0.});
+    auto R2 = RotationVectorToQuaternion({0., 1.0, 0.});
```
faisal-bhuiyan, line 176: For consistency

```
    auto R1 = RotationVectorToQuaternion({1., 0., 0.});
    auto R2 = RotationVectorToQuaternion({0., 1., 0.});
```

tests/unit_tests/system/test_calculate_RR0.cpp
```diff
@@ -8,28 +8,23 @@
...
-    constexpr auto expected_rr0_data = std::array{
-        2780., 4400.,  -400., 0.,     0.,    0.,    -3280., 2372., 3296., 0.,    0.,     0.,
-        2960., -1504., 4028., 0.,     0.,    0.,    0.,     0.,    0.,    2780., 4400.,  -400.,
-        0.,    0.,     0.,    -3280., 2372., 3296., 0.,     0.,    0.,    2960., -1504., 4028.};
+    constexpr auto expected_rr0_data = std::array{-44., 4.,   118., 0.,   0.,   0.,    //
+                                                  116., -22., 44.,  0.,   0.,   0.,    //
+                                                  22.,  124., 4.,   0.,   0.,   0.,    //
+                                                  0.,   0.,   0.,   -44., 4.,   118.,  //
+                                                  0.,   0.,   0.,   116., -22., 44.,   //
+                                                  0.,   0.,   0.,   22.,  124., 4.};
```
faisal-bhuiyan, line 27: Why did these values change?

faisal-bhuiyan (approved):
> This PR adds several kernels to calculate the current node and quadrature point locations, which is needed for the following reasons
> 
> * VTK outputs
> * For any future output data writing
> * Interfacing with the `AeroDyn-Inflow` external library for calculating aerodynamic loads because the positions of the blade node/quadrature points will need to be passed on


Given all of the above are optional additions to enhance OpenTurbine, I wonder about the design to add the global position and rotation to the `State` class itself. Could it be more flexible and modular to create a separate class to handle this responsibility if requested?

src/beams/beams.hpp
```diff
@@ -54,6 +55,7 @@ struct Beams {
     Kokkos::View<double** [4]> qp_r_prime;             // State: rotation derivative
     Kokkos::View<double** [3]> qp_omega;               // State: angular velocity
     Kokkos::View<double** [3]> qp_omega_dot;           // State: position/rotation
+    Kokkos::View<double** [3]> qp_deformation;         // Deformation relative to undeformed state
```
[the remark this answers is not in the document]
deslaughter: How about `// Deformation relative to rigid body motion`

src/beams/calculate_QP_deformation.hpp
```diff
@@ -0,0 +1,84 @@
...
+        auto x_root = View_3::const_type(x_root_data.data());
+
+        // Element root node rotational displacement
+        auto r_root_data = Kokkos::Array<double, 4>{
+            qp_r_(i_elem, 0, 0),
+            qp_r_(i_elem, 0, 1),
+            qp_r_(i_elem, 0, 2),
+            qp_r_(i_elem, 0, 3),
+        };
+        auto r_root = View_Quaternion::const_type(r_root_data.data());
```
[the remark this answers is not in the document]
deslaughter: `x0_root`, `x_root`, and `r_root` are all root based data which I think should go together. It's easier to check the indexing that way.

src/model/model.hpp
deslaughter: It should be moving it. Is there an annotation needed for that to happen?

tests/unit_tests/system/test_calculate_RR0.cpp
deslaughter: Instead of specifying `R` and `R0`, the new test specifies `RR0`, which is the current global rotation, as a quaternion, which is then turned into the `RR0` matrix. This was done so that the global rotation could be calculated in `calculate_QP_position` and stored for later use. The values changed because the inputs changed.

PR #274, opened 2024-09-18
Fuse node loops for matrix integration, exposing more parallelism on GPUs

ddement:
The result is a 4-5x speed up of those kernels on GPU and a 20%+ speed up overall of the 10s IEA15RotorHub problem.

This potentially creates a pattern that we can use for other middle loops that can be SIMD-ified or else have more parallelism exposed for GPU work.

### DIFF
--- a/src/system/integrate_inertia_matrix.hpp
+++ b/src/system/integrate_inertia_matrix.hpp
@@ -19,49 +19,52 @@ struct IntegrateInertiaMatrixElement {
     Kokkos::View<double*** [6][6]> gbl_M_;
 
     KOKKOS_FUNCTION
-    void operator()(size_t i_index) const {
+    void operator()(size_t ij_index) const {
         using simd_type = Kokkos::Experimental::native_simd<double>;
         using mask_type = Kokkos::Experimental::native_simd_mask<double>;
         using tag_type = Kokkos::Experimental::element_aligned_tag;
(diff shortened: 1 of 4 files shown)


src/system/integrate_inertia_matrix.hpp
```diff
@@ -19,49 +19,52 @@ struct IntegrateInertiaMatrixElement {
     Kokkos::View<double*** [6][6]> gbl_M_;
 
     KOKKOS_FUNCTION
-    void operator()(size_t i_index) const {
+    void operator()(size_t ij_index) const {
         using simd_type = Kokkos::Experimental::native_simd<double>;
```
deslaughter, line 23: Production software should not depend on experimental library features.

ddement: This commit doesn't add the dependence on Kokkos SIMD (this has already been in the code for the matrix integration loops)

With regards to the experimental features, to its spot as the basis for a lot of codes, Kokkos is pretty conservative about its "Experimental" tag.  This gives them a more freedom to add to or change the API between major annual releases, where non-experimental code has very tight stability requirements.  They have even maintained backwards compatibility when moving things out of the Experimental namespace.  The code itself is extensively tested and documented.  Our usage is very vanilla and idiomatic, so it is unlikely to ever need to change substantially.

deslaughter (approved):

PR #276, opened 2024-09-24
Fixes some new warnings (perhaps from CppCheck?) exposed during development

ddement:


deslaughter:
I'm sure this is a style preference, but what does CppCheck have against `for` loops? I understand the issues with potentially changing the iterator variable, but requiring `std::transform_reduce` seems extreme. Future developers, and me, are probably going to struggle with this.

deslaughter (approved):

PR #279, opened 2024-09-25
Split functions within the step folder into .hpp and .cpp files

ddement:
This commit should improve compile times for incremental builds quite significantly, depending on where you're making your changes.  Note that it seems to slow down building from scratch by a little bit (about 20-50%).  

The functions in this folder mark the transition from a high level "pass around the structs" workflow to actually calling the kernels that do the work, so they are a reasonable place to provide the split to get the most "bang for our buck" with incremental compile time improvements.


deslaughter (changes requested):
I don't think forward declaring the structs in the header files is a good idea. I understand that it allows the headers to be decoupled, but then if you want to actually use those headers you have to track down the origin of each type that has been forward declared. This makes the library harder to use and is likely to frustrate the users. I'd recommend going ahead and referencing the appropriate includes.

ddement:
I've added the top-level `openturbine.hpp` for users to include rather than being fine-grained with their includes.  I'm open to adding more to this, but I figured this was a good first set of files and the bare minimum for using OpenTurbine.  

Let me know if this looks good or if you'd still prefer to include in the lower level headers.

src/openturbine.hpp
faisal-bhuiyan, line 1: What is the intent for this file?

src/step/assemble_constraints_matrix.hpp
```diff
@@ -1,58 +1,8 @@
...
-        );
-        Kokkos::deep_copy(tmp_row_map, solver.B_t.graph.row_map);
-        Kokkos::parallel_for(
-            "CopySparseValuesToTranspose", constraint_transpose_policy,
-            CopySparseValuesToTranspose<Solver::CrsMatrixType>{solver.B, tmp_row_map, solver.B_t}
-        );
-        KokkosSparse::sort_crs_matrix(solver.B_t);
-    }
+struct Solver;
+struct Constraints;
```
faisal-bhuiyan, line 5: This makes me nervous -- why are we using forward declarations here? I think the risks outweigh the compile-time savings here (e.g. unit testing this -- would there be surprises?). For what it's worth, Google's style guild also discourages it.

https://google.github.io/styleguide/cppguide.html#Forward_Declarations

src/step/assemble_constraints_matrix.hpp
```diff
@@ -1,58 +1,8 @@
...
 
-    {
-        auto mult_region = Kokkos::Profiling::ScopedRegion("Assemble Constraints Matrix");
-        KokkosSparse::spgemm_numeric(
-            solver.constraints_spgemm_handle, solver.B, false, solver.T, false,
-            solver.constraints_matrix
-        );
-    }
-}
+void AssembleConstraintsMatrix(Solver& solver, const Constraints& constraints);
```
faisal-bhuiyan, line 7: Do we have a policy on declaring parameter names (which are not primitive) in the headers-only files? For example, would the following be cleaner?

`void AssembleConstraintsMatrix(Solver&, const Constraints&);`

faisal-bhuiyan:
> The functions in this folder mark the transition from a high level "pass around the structs" workflow to actually calling the kernels that do the work, so they are a reasonable place to provide the split to get the most "bang for our buck" with incremental compile time improvements.

Just wondering if any other parts of openturbine might benefit from such a split -- what are your thoughts? Is maintaining a mix of header-only files + splitting them out as appropriate the optimal path?

src/openturbine.hpp
ddement: This file is to provide a convenience header for users to include rather than having to know which ones they really want.  For example, if they want our step function and all of the structs that it depends on, they would otherwise have to include 5 different headers.  Before this change, they would have gotten them all from including `step.hpp`, but this is violating the guideline of including what you use and sets them up for breakage if we make any changes.  

Internally, we should not use this header and continue including everything we want directly.

src/step/assemble_constraints_matrix.hpp
ddement: I don't think we have a policy for this yet.  I don't have a strong opinion about which I prefer.

ddement:
> > The functions in this folder mark the transition from a high level "pass around the structs" workflow to actually calling the kernels that do the work, so they are a reasonable place to provide the split to get the most "bang for our buck" with incremental compile time improvements.
> 
> Just wondering if any other parts of openturbine might benefit from such a split -- what are your thoughts? Is maintaining a mix of header-only files + splitting them out as appropriate the optimal path?

Maybe, but I don't think so.  This split already slows down our compile times from scratch significantly.  Truth be told, I'm not even sure if this change is a good idea (I put it together hoping it would be a massive, unqualified win).  If we ever decide to do any templating on these functions, we'll have to pull them back into header files, for example.  

This break location was also chosen because it has the best chance of insulating our users from long compile times (they will mostly use the functions in the `step` folder).  Given our discussions in recent weeks about how we want our users to interact with OpenTurbine, it may make sense to compile to objects at that higher level interface and make most things below it be headers.

src/step/assemble_constraints_matrix.hpp
ddement: Forward declarations are good for decreasing compile times and binary size, but are also for decoupling our code and ensuring that we are doing a good job of managing our dependencies.  

an example: in our unit tests, we created and manipulated the `Constraints` and `StepParameters` structs.  However, we never actually included them, instead getting their code transitively from `Step.hpp`.  A change in our code layout broke our tests and could break user code.  

Their cautionary tales against the forward declaration seem to pertain to using a forward declaration within an implementation.  I agree that this is bad.  I don't think they apply when using forward declarations when defining your interface, which is common practice.

PR #289, opened 2024-10-15
Changed CMake build to work when Trilinos is built with MPI enabled

ddement:
This is the default for Spack builds and will likely be how Trilinos is configured when we're used with large scale projects (i.e. Nalu-Wind).  We will be able to link against a such a Trilinos configuration instead of requiring our own non-MPI version.


src/CMakeLists.txt
```diff
@@ -16,13 +16,20 @@ target_link_libraries(openturbine_library PRIVATE
 )
 
 # Link system libraries
-target_link_system_libraries(openturbine_library PRIVATE
+target_link_system_libraries(openturbine_library PUBLIC
```
faisal-bhuiyan, line 19: Good catch

.github/workflows/correctness-linux.yaml
```diff
@@ -31,7 +31,7 @@ jobs:
         spack install googletest
         spack install cppcheck
         spack install yaml-cpp
-        spack install trilinos@master~mpi~epetra
+        spack install trilinos@master~epetra
```
faisal-bhuiyan, line 34: I thought we could get away without putting `mpi` into CI -- I suppose it's a check for the code?

faisal-bhuiyan (approved):
Looks good

.github/workflows/correctness-linux.yaml
ddement: Yes, this is just to check to ensure it continues to work.  The MacOS version still tests with a version of Trilinos built without MPI, so both cases are covered. 

In truth, even Linux CI is still using the old cache right now, but it will build a new Trilinos with MPI next time we rebuild it.

PR #292, opened 2024-10-22
Build recipes for openturbine

faisal-bhuiyan:
This pull request introduces a `recipes` directory with two introductory formulas to streamline the build process for OpenTurbine: a build recipe for macOS (`build_recipe_macos.sh`) and a `Dockerfile` for containerized builds. 

- The macOS script automates the installation of dependencies using `Spack`, sets up the build environment, and compiles OpenTurbine with configurable options
- The Dockerfile creates a Ubuntu-based container that installs necessary tools, uses `Spack` to manage dependencies, and builds OpenTurbine with sanitizers and additional features enabled.

These additions aim to lower the barrier to entry for contributing to OpenTurbine, making it easier for developers to set up and contribute to the project.

### DIFF
--- a/recipes/Dockerfile
+++ b/recipes/Dockerfile
@@ -1,3 +1,6 @@
+# Run this from the root of the OpenTurbine repository e.g.
+# docker build -f recipes/Dockerfile -t openturbine:latest .
+
 FROM ubuntu:latest
 
 # Use bash as the default shell
--- a/recipes/Dockerfile
+++ b/recipes/Dockerfile
@@ -17,6 +17,7 @@ RUN apt-get update -qq && \
         make \
         python3 \
         python3-pip \
+        unzip \
         wget && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
@@ -48,11 +49,12 @@ ARG CMAKE_OPTIONS="-DOpenTurbine_ENABLE_SANITIZER_ADDRESS=ON \
     -DOpenTurbine_BUILD_ROSCO_CONTROLLER=ON \
     -DCMAKE_BUILD_TYPE=Debug"
 
+ARG N_PARALLEL_JOBS=4
+
 RUN . /spack/share/spack/setup-env.sh && \
-    spack load trilinos && \
-    spack load yaml-cpp && \
-    spack load vtk && \
+    spack load cppcheck googletest trilinos vtk yaml-cpp && \
     mkdir docker-build && \
     cd docker-build && \
     cmake .. ${CMAKE_OPTIONS} && \
-    cmake --build . -j 4
+    cmake --build . -j ${N_PARALLEL_JOBS} && \
+    ctest --output-on-failure -j ${N_PARALLEL_JOBS}
--- a/recipes/Dockerfile
+++ b/recipes/Dockerfile
@@ -1,0 +1,58 @@
+FROM ubuntu:latest
+
+# Use bash as the default shell
+SHELL ["/bin/bash", "-c"]
+
+# Step 1: Install required packages and tools
+RUN apt-get update -qq && \
+    apt-get install -y \
+        build-essential \
+        clang-tidy \
+        cmake \
+        curl \
+        g++ \
+        gfortran \
+        git \
+        libcurl4-openssl-dev \
+        make \
+        python3 \
+        python3-pip \
+        wget && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+# Step 2: Get spack and install necessary libraries using it
+RUN git clone https://github.com/spack/spack.git && \
+    source spack/share/spack/setup-env.sh && \
+    spack compiler find && \
+    spack external find && \
+    spack mirror add spack-public https://mirror.spack.io && \
+    spack install googletest && \
+    spack --insecure install cppcheck && \
+    spack install yaml-cpp && \
+    spack --insecure install trilinos@16.0.0~mpi~epetra && \
+    spack --insecure install vtk~mpi~opengl2
+
+# Step 3: Set up the working directory and copy the OpenTurbine repository
+WORKDIR /projects/openturbine
+COPY . .
+
+# Step 4: Load necessary libraries and build the project
+ARG CMAKE_OPTIONS="-DOpenTurbine_ENABLE_SANITIZER_ADDRESS=ON \
+    -DOpenTurbine_ENABLE_SANITIZER_LEAK=ON \
+    -DOpenTurbine_ENABLE_SANITIZER_UNDEFINED=ON \
+    -DOpenTurbine_ENABLE_CPPCHECK=ON \
+    -DOpenTurbine_ENABLE_CLANG_TIDY=ON \
+    -DOpenTurbine_ENABLE_VTK=ON \
+    -DOpenTurbine_BUILD_OPENFAST_ADI=ON \
+    -DOpenTurbine_BUILD_ROSCO_CONTROLLER=ON \
+    -DCMAKE_BUILD_TYPE=Debug"
+
+RUN . /spack/share/spack/setup-env.sh && \
+    spack load trilinos && \
+    spack load yaml-cpp && \
+    spack load vtk && \
+    mkdir docker-build && \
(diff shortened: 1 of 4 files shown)


faisal-bhuiyan:
@ddement Do you think a recipe for GPU builds might be worth adding here?

recipes/build_recipe_macos.sh
```diff
@@ -0,0 +1,80 @@
...
+    if ! spack find -l "$1" | grep -q "$1"; then
+        echo "Installing $1..."
+        spack install "$1"
+    else
+        echo "$1 is already installed."
+    fi
+}
+
+install_if_missing googletest
+install_if_missing llvm
```
ddement, line 38: This is only needed for clang-tidy.  Let's save our users some time and not include it in this file.

recipes/build_recipe_macos.sh
```diff
@@ -0,0 +1,80 @@
...
+spack_path=$(dirname "$2") # Get the directory of the provided path
+if [ ! -d "$spack_path/spack" ]; then
+    git clone https://github.com/spack/spack.git $spack_path/spack
+fi
+source $spack_path/spack/share/spack/setup-env.sh
+spack compiler find
+spack external find
+
+# Install required packages using Spack if not present
+install_if_missing() {
```
ddement, line 28: I thought `spack install` would issue a similar message if the requested spec was already installed.  I don't think this wrapper should be necessary.

recipes/build_recipe_macos.sh
```diff
@@ -0,0 +1,80 @@
...
+        echo "Installing $1..."
+        spack install "$1"
+    else
+        echo "$1 is already installed."
+    fi
+}
+
+install_if_missing googletest
+install_if_missing llvm
+install_if_missing cppcheck
```
ddement, line 39: same as with llvm

recipes/build_recipe_macos.sh
```diff
@@ -0,0 +1,80 @@
...
+spack load googletest
+spack load yaml-cpp
+spack load vtk
+
+# Find gfortran compiler
+if ! command -v brew &> /dev/null; then
+    echo "Homebrew is not installed. Please install it first."
+    exit 1
+fi
+export FC=$(find /opt/homebrew/ -name gfortran | tr ' ' '\n' | grep "/gcc/.*gfortran")
```
ddement, line 57: I think we should just check that the user has defined FC in their environment and tell them to do so if they have not.  They may have a preferred Fortan compiler that they want to use.

recipes/build_recipe_macos.sh
```diff
@@ -0,0 +1,80 @@
...
+cmake .. \
+  -DOpenTurbine_ENABLE_SANITIZER_ADDRESS=$([[ "$BUILD_EXTERNAL" == "none" ]] && echo "ON" || echo "OFF") \
+  -DOpenTurbine_ENABLE_SANITIZER_UNDEFINED=$([[ "$BUILD_EXTERNAL" == "none" ]] && echo "ON" || echo "OFF") \
+  -DOpenTurbine_ENABLE_CPPCHECK=ON \
+  -DOpenTurbine_ENABLE_CLANG_TIDY=ON \
+  -DOpenTurbine_ENABLE_VTK=ON \
+  -DOpenTurbine_BUILD_OPENFAST_ADI=$([[ "$BUILD_EXTERNAL" == "all" ]] && echo "ON" || echo "OFF") \
+  -DOpenTurbine_BUILD_ROSCO_CONTROLLER=$([[ "$BUILD_EXTERNAL" == "all" ]] && echo "ON" || echo "OFF") \
+  -DCMAKE_CXX_COMPILER="$CXX_COMPILER" \
+  -DCMAKE_BUILD_TYPE="$BUILD_TYPE"
```
ddement, line 75: Since we're building release, let's give them a good production environment to work with:
```
cmake .. \
  -DOpenTurbine_ENABLE_VTK=ON \
  -DOpenTurbine_BUILD_OPENFAST_ADI=ON \
  -DOpenTurbine_BUILD_ROSCO_CONTROLLER=ON
  -DCMAKE_BUILD_TYPE=Release
```

recipes/build_recipe_macos.sh
```diff
@@ -0,0 +1,80 @@
...
+if ! command -v brew &> /dev/null; then
+    echo "Homebrew is not installed. Please install it first."
+    exit 1
+fi
+export FC=$(find /opt/homebrew/ -name gfortran | tr ' ' '\n' | grep "/gcc/.*gfortran")
+
+# Build OpenTurbine with following options - modify as needed
+BUILD_TYPE="Release" # Release, Debug, RelWithDebInfo
+CXX_COMPILER="clang++" # g++, clang++
+BUILD_EXTERNAL="all" # all, none
```
ddement, line 62: I don't think we need to define these here (see my proposed alternative cmake command)

recipes/build_recipe_macos.sh
```diff
@@ -0,0 +1,80 @@
...
+  -DOpenTurbine_ENABLE_VTK=ON \
+  -DOpenTurbine_BUILD_OPENFAST_ADI=$([[ "$BUILD_EXTERNAL" == "all" ]] && echo "ON" || echo "OFF") \
+  -DOpenTurbine_BUILD_ROSCO_CONTROLLER=$([[ "$BUILD_EXTERNAL" == "all" ]] && echo "ON" || echo "OFF") \
+  -DCMAKE_CXX_COMPILER="$CXX_COMPILER" \
+  -DCMAKE_BUILD_TYPE="$BUILD_TYPE"
+
+cmake --build .
+
+# Run tests
+ctest --output-on-failure
```
ddement, line 80: No need to run the tests as part of a build script.

recipes/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+        gfortran \
+        git \
+        libcurl4-openssl-dev \
+        make \
+        python3 \
+        python3-pip \
+        unzip \
+        wget && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
```
ddement, line 26: Do we have permissions to do this in Docker?  These commands typically would require `sudo`.

recipes/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+    spack install yaml-cpp && \
+    spack --insecure install trilinos@16.0.0~mpi~epetra && \
+    spack --insecure install vtk~mpi~opengl2
+
+# Step 3: Set up the working directory and copy the OpenTurbine repository
+WORKDIR /projects/openturbine
+COPY . .
+
+# Step 4: Load necessary libraries and build the project
+ARG CMAKE_OPTIONS="-DOpenTurbine_ENABLE_SANITIZER_ADDRESS=ON \
```
ddement, line 45: I think the cmake options should mirror what I proposed below for the MacOS script - something production ready for a non-power user to get a build that is ready to deploy.

ddement (commented):
Great idea to put this together.  Once these are in main, I'll create versions for HIP and CUDA.

recipes/build_recipe_macos.sh
faisal-bhuiyan: Sounds good -- let's comment it out and provide it as an option. I'll do the same for the vtk module.

recipes/build_recipe_macos.sh
faisal-bhuiyan: It does not really issue any messages, so I added this wrapper just to make sure we're not duplicating anything there. However, it does not try to re-install it either if the package is found.

recipes/build_recipe_macos.sh
faisal-bhuiyan: I agree, let's provide a minimal set for the default

recipes/build_recipe_macos.sh
faisal-bhuiyan: Sounds good to me

recipes/Dockerfile
faisal-bhuiyan: I don't think you need `sudo` for this.

ddement (approved):

PR #293, opened 2024-10-22
Move Math and Parser tests out of regression tests

ddement:
This is the first step before I move our true regression tests to their own executable.


faisal-bhuiyan (approved):
Looks good (added couple of minor commits as I was reviewing this)

PR #300, opened 2024-11-05
Added initial data structures and algorithms to support different element types

ddement:
This commit adds the element freedom signature, element freedom table, node freedom allocation table, and node freedom map tables necessary for supporting nodes with different numbers of degrees of freedom.  It also adds the free functions necessary for populating them on the State and Beams classes.  The algorithms are currently serial for ease in refactoring as we add features, but can easily make them parallel in the future.

### DIFF
--- a/src/dof_management/freedom_signature.hpp
+++ b/src/dof_management/freedom_signature.hpp
@@ -7,9 +7,9 @@
 namespace openturbine {
 
 enum class FreedomSignature : std::uint8_t {
-    AllComponents = 0b01111111,
-    JustPosition = 0b01110000,
-    JustRotation = 0b00001111,
+    AllComponents = 0b00111111,
+    JustPosition = 0b00111000,
+    JustRotation = 0b00000111,
     NoComponents = 0b00000000
 };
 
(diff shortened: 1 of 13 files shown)


src/dof_management/freedom_signature.hpp
```diff
@@ -0,0 +1,34 @@
...
+    return static_cast<FreedomSignature>(static_cast<T>(x) | static_cast<T>(y));
+}
+
+KOKKOS_INLINE_FUNCTION
+size_t count_active_dofs(FreedomSignature x) {
+    using T = std::underlying_type_t<FreedomSignature>;
+    auto count = 0UL;
+    constexpr auto zero = T{0};
+    constexpr auto one = T{1};
+    for (auto value = static_cast<T>(x); value > zero; value = value >> 1) {
```
deslaughter, line 28: It's a shame that https://en.cppreference.com/w/cpp/numeric/popcount is in C++20, it would be great here.

deslaughter (approved):

src/dof_management/freedom_signature.hpp
ddement: Agreed. I initially prototyped this this bitset's count function, which does the same thing. Sadly, we can't expect to be able to use the standard library on device, so we have to roll our own.

PR #304, opened 2024-11-11
Least-Squares fitting for high-order beam blade models

faisal-bhuiyan:
Implements https://github.com/Exawind/openturbine-project/issues/5

PS: There is a TODO item to implement a dense linear solver, which should probably be a separate task.


ddement (approved):
Looks good for small problems.  Once we have the linear solver in place some real use cases, we should benchmark it to see if there's anything that needs to be optimized - in particular, using a `std::vector<std::vector<double>>` is really slow in general.

deslaughter (approved):

PR #312, opened 2024-11-26
Implement linear solver for least squares

ddement:
This changes our TPL requirements.  We now require our Kokkos-Kernels to be configured to use Lapack, which gives us access to Lapack's gesv function under the hood.


.github/workflows/correctness-linux.yaml
```diff
@@ -31,7 +31,7 @@ jobs:
         spack install googletest
         spack install cppcheck
         spack install yaml-cpp
-        spack install trilinos@master~epetra
+        spack install trilinos@master~epetra ^kokkos-kernels+blas+lapack
```
faisal-bhuiyan, line 34: Could you add this to the build recipes?

src/beams/interpolation.hpp
```diff
@@ -166,7 +166,8 @@ inline std::vector<double> GenerateGLLPoints(const size_t order) {
     }
 
     const size_t n_nodes = order + 1;
-    std::vector<double> gll_points(n_nodes, 0.);
+    auto gll_points = std::vector<double>(n_nodes, 0.);
+    gll_points.resize(n_nodes);
```
faisal-bhuiyan, line 170: Is this resizing necessary here, since we already initialize the `std::vector`?

faisal-bhuiyan (approved):
Looks good

deslaughter (approved):

.github/workflows/correctness-linux.yaml
ddement: Will do

src/beams/interpolation.hpp
ddement: It shouldn't be.  In fact, it's a no-op.  I was getting compiler warnings locally (gcc 12.3) from the `.front()` and `.back()` calls saying that it was dereferencing a null pointer.  Obviously, the warnings were incorrect, but calling resize here silences them.  I wish I knew a better way.

PR #315, opened 2024-12-04
Add utilities for copying the State data structure

ddement:
This PR adds two new utility functions: CloneState and CopyStateData.

CloneState allocates memory for a new state and copies all data from an existing one.  This should be used in problem initialization and results in a fully usable clone of the existing state (i.e. `auto archive_state = CloneState(current_state)`.  

CopyStateData does not allocate memory and only copies that data which will change over the course of a time step.  Data which will not change during a time step is not copied as a performance optimization.  This should be used during time stepping when a state wants to be saved (i.e. `CopyStateData(archive_state, current_state)`) or restored (i.e. `CopyStateData(current_state, archive_state)`).

### DIFF
--- a/src/state/clone_state.hpp
+++ b/src/state/clone_state.hpp
@@ -1,0 +1,24 @@
+#pragma once
+
+#include "state.hpp"
+
+namespace openturbine {
+
+inline State CloneState(const State& old) {
+    auto clone = State(old.num_system_nodes);
+    Kokkos::deep_copy(clone.ID, old.ID);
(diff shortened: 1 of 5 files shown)


deslaughter (approved):
Will Kokkos throw an error in `CopyStateData` if the view sizes are different between `copy` and `old`?

ddement:
> Will Kokkos throw an error in `CopyStateData` if the view sizes are different between `copy` and `old`?

@deslaughter 
Yes.  I just put together a quick test of deep copy with differently sized views, and it throws an exception with this error:
 "Deprecation Error: Kokkos::deep_copy extents of views don't match: A(3,4) b(4,3) "

PR #321, opened 2024-12-14
Implement R3-based geometrically nonlinear, constitutively linear spring element

faisal-bhuiyan:
Implements the task described [here](https://github.com/Exawind/openturbine-project/issues/19#issue-2715956703). The system kernels were implemented here -- leaving the assembly logic to be implemented in a follow up work.


deslaughter:
I think it looks pretty good. Most of the CI errors are just cleanup.

tests/regression_tests/regression/test_springs.cpp
```diff
@@ -0,0 +1,58 @@
...
+        1.      // Undeformed length
+    )});
+
+    return CreateSprings(springs_input);
+}
+
+TEST(SpringsTest, NodeStateIndices) {
+    const auto springs = SetUpSprings();
+    EXPECT_EQ(springs.node_state_indices(0, 0), 0);
+    EXPECT_EQ(springs.node_state_indices(0, 1), 1);
```
ddement, line 35: These checks access Device data on host and segfault for GPU builds.  Copy them down to host first.

ddement:
Looks good to me other than the GPU segfault mentioned above

tests/regression_tests/regression/test_springs.cpp
faisal-bhuiyan: Could you confirm if the changes fixed the segfault?

ddement: It works now!

ddement (approved):

PR #322, opened 2024-12-18
Added unit tests for the kernels computing the system variables for masses

ddement:


faisal-bhuiyan (approved):
Looks good -- all these are adopted from the Beams folder?

ddement:
> Looks good -- all these are adopted from the Beams folder?

Yes - I did it that way to ensure that I didn't accidentally introduce any changes to the math between the two implementations.

PR #329, opened 2025-01-10
Added a regression test with a chain of masses/springs which are in stationary equilibrium.

ddement:
This test is a more complex configuration with a very simple analytic solution.  More importantly, it is parametrically scalable, so it can be easily scaled up/down for performance testing.

### DIFF
--- a/tests/regression_tests/regression/test_spring_mass_system.cpp
+++ b/tests/regression_tests/regression/test_spring_mass_system.cpp
@@ -124,4 +124,118 @@ TEST(SpringMassSystemTest, FinalDisplacement) {
     SetUpSpringMassSystem();
 }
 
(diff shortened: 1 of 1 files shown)


tests/regression_tests/regression/test_spring_mass_system.cpp
```diff
@@ -124,4 +124,118 @@ TEST(SpringMassSystemTest, FinalDisplacement) {
     SetUpSpringMassSystem();
 }
 
+inline auto SetUpSpringMassChainSystem() {
+    auto model = Model();
+
+    // Add two nodes for the spring element
```
faisal-bhuiyan, line 130: Let's remove or update the comments throughout the test.

PR #332, opened 2025-01-14
Added CreateSystem and CreateSystemWithSolver methods to model.

ddement:
These methods give users a shorter way to create the structures needed to use OpenTurbine.  CreateSystem just creates the State, Elements, and Constraints, while CreateSystemWithSolver also creates the Solver.

To support this, a copy constructor was added to Solver.  This is needed because the default copy constructor was not properly initializing the new Kernel Handles needed for sparse matrix addition and multiplication.


deslaughter:
Thanks for getting these working, it makes the API look a lot nicer for the user.

deslaughter (approved):

PR #335, opened 2025-01-16
Add Install capabilities to CMake to support distribution

ddement:
This adds the necessary statements to CMake so that `make install` will install our library file and all header files  It also provides the necessary files for operating with CMake's find_package command.


deslaughter (approved):
I like that `src` is no longer in all the header file include paths. 

Could we add at test that uses the library? Or is there a better way to check that new files are added to the `install(FILES` commands? We've gotten burned by this OpenFAST where new things are added and the code is released and we don't find out that a file is missing from the install until after the downstream users try to use it.

ddement:
> I like that `src` is no longer in all the header file include paths.
> 
> Could we add at test that uses the library? Or is there a better way to check that new files are added to the `install(FILES` commands? We've gotten burned by this OpenFAST where new things are added and the code is released and we don't find out that a file is missing from the install until after the downstream users try to use it.

Yes - I've been thinking about how to do that.  I think I can put something together using CI whenever we push to main.  My current plan is add a line which installs OpenTurbine (and all of its dependencies) through Spack and then builds an application linking to it.  I will add that once we have our Spack package submitted to that repo.

PR #347, opened 2025-02-04
Added restart capability for CFD Interface.

ddement:
It currently only requires reading in the state data, as the other parameters needed to create the initial system should be readily available to calling code through their input files.  It also gives the ability to make changes (such as adding quadrature points or changing timestep parameters) inbetween restarts.


deslaughter (commented):
Could you put the main part of this functionality outside of the interface, maybe as a function in state? If we have multiple interfaces that need to implement a save/restore then I don't want to re-implement this for each one.

deslaughter (approved):

PR #348, opened 2025-02-05
Quaternion from tangent and twist

deslaughter:
Add function to compute quaternion from material axis tangent and twist parameter.

### DIFF
--- a/src/math/quaternion_operations.hpp
+++ b/src/math/quaternion_operations.hpp
@@ -66,7 +66,7 @@ inline Array_4 RotationMatrixToQuaternion(const Array_3x3& m) {
     };
 
     // Get maximum value and index of maximum value
-    auto max_num = std::max_element(vals.begin(), vals.end());
+    auto* max_num = std::max_element(vals.begin(), vals.end());
     auto max_idx = std::distance(vals.begin(), max_num);
 
     auto tmp = sqrt(*max_num + 1.);
--- a/src/math/quaternion_operations.hpp
+++ b/src/math/quaternion_operations.hpp
@@ -51,7 +51,9 @@ inline std::array<Array_3, 3> QuaternionToRotationMatrix(const Array_4& q) {
 }
 
 /**
- * @brief Converts a 3x3 rotation matrix to a 4x1 quaternion and returns the result
+ * @brief Converts a 3x3 rotation matrix to a 4x1 quaternion and returns the result, see
+ * https://www.euclideanspace.com/maths/geometry/rotations/conversions/matrixToQuaternion/ for
+ * implementation details
  */
 inline Array_4 RotationMatrixToQuaternion(const Array_3x3& m) {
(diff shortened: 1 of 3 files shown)


tests/unit_tests/math/test_quaternion_operations.cpp
```diff
@@ -380,4 +380,20 @@ TEST(QuaternionTest, QuaternionToRotationVector_2) {
     }
 }
```
faisal-bhuiyan, line 382: Would it make sense to add a unit test for the `RotationMatrixToQuaternion` function separately?

src/math/quaternion_operations.hpp
```diff
@@ -245,4 +305,30 @@ Kokkos::Array<double, 4> NormalizeQuaternion(const Kokkos::Array<double, 4>& q)
...
 
+/**
+ * @brief Returns a 4-D quaternion from provided tangent vector and twist (degrees) about tangent
+ */
+inline Array_4 TangentTwistToQuaternion(const Array_3& tangent, const double twist) {
+    const auto e1 = UnitVector(tangent);
+    auto a = e1[0] > 0. ? 1. : -1.;
+    const Array_3 e2{
+        -a * e1[1] / sqrt(e1[0] * e1[0] + e1[1] * e1[1]),
+        a * e1[0] / sqrt(e1[0] * e1[0] + e1[1] * e1[1]),
```
faisal-bhuiyan, line 316: Couple comments
- This returns `NaN` when the tangent is the z-axis, i.e. `e1 = e2 = 0.` -- we should probably check for div by zero here
- We could simplify this by introducing a local variable as such
```
    const auto denominator = std::sqrt(e1[0] * e1[0] + e1[1] * e1[1]);
    const Array_3 e2{
        -a * e1[1] / denominator,
        a * e1[0] / denominator,
        0.,
    };
```

deslaughter:
I think that I've addressed the comments, please take another look.

tests/unit_tests/math/test_quaternion_operations.cpp
deslaughter: Added a test for this

src/math/quaternion_operations.hpp
```diff
@@ -49,6 +50,65 @@ inline std::array<Array_3, 3> QuaternionToRotationMatrix(const Array_4& q) {
...
+    }
+    if (max_idx == 3) {
+        return Array_4{
+            (m[1][0] - m[0][1]) * c,
+            (m[0][2] + m[2][0]) * c,
+            (m[1][2] + m[2][1]) * c,
+            0.5 * tmp,
+        };
+    }
+    return Array_4{1., 0., 0., 0};
```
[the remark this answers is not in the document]
deslaughter: Added a link to website that describes the algorithm

faisal-bhuiyan (approved):

PR #349, opened 2025-02-05
Refactor LSFE logic to add derivative of shape functions

faisal-bhuiyan:
Quick patch to the LSFE logic to add the shape function derivatives.

### DIFF
--- a/tests/unit_tests/math/test_least_squares_fit.cpp
+++ b/tests/unit_tests/math/test_least_squares_fit.cpp
@@ -51,7 +51,7 @@ TEST(LeastSquaresFitTest, ShapeFunctionMatrices_FirstOrder) {
     const size_t n{3};                               // Number of pts to fit
     const size_t p{2};                               // Polynomial order + 1
     const std::vector<double> xi_g = {-1., 0., 1.};  // Evaluation points
-    const auto [phi_g, gll_pts] = openturbine::ShapeFunctionMatrices(n, p, xi_g);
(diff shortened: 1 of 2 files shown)


tests/unit_tests/math/test_least_squares_fit.cpp
```diff
@@ -74,13 +74,30 @@ TEST(LeastSquaresFitTest, ShapeFunctionMatrices_FirstOrder) {
...
+
+    // Check derivative shape function matrix dimensions (2 x 3)
+    ASSERT_EQ(dphi_g.size(), p);
+    ASSERT_EQ(dphi_g[0].size(), n);
+    ASSERT_EQ(dphi_g[1].size(), n);
+
+    // Check derivative shape function values at evaluation points
+    const std::vector<std::vector<double>> expected_dphi_g = {
+        {-0.5, -0.5, -0.5},  // First derivative shape function
+        {0.5, 0.5, 0.5}      // Second derivative shape function
```
deslaughter, line 86: These comments are incorrect. The shape function derivative weights are organized by column, not row; and it's `first shape function derivative`, not first `derivative of shape function`. Did Copilot write these comments?

faisal-bhuiyan: Sorry for the rather clumsy comments -- those were wrong in the original shape function checks and I wrote the commit pretty late into the evening to catch it. Will be more careful with this before submitting a PR for review.

deslaughter (approved):

PR #351, opened 2025-02-07
Add trapezoidal quadrature

deslaughter:
Added a function and associated tests to create a trapezoidal quadrature from a series of grid locations. This will be used to generate quadrature corresponding to beam section property locations for the blade interface.


ddement (approved):
Looks good. We should proliferate this to our regression tests that create their trapezoidal quadrature manually.

PR #352, opened 2025-02-10
Added InterfaceBuilder to improve ergonomics of creating our CFD interface

ddement:


deslaughter (commented):
I like the overall concept and it better documents the values that the user species.

My main concern is the ability of a user to discover which fields must be set. That will need to be documented or the builder will need to keep track of which values have been set and throw an error on the final build call after checking all the inputs. I found there was also something called a _Step Builder_ which uses additional interfaces to expose different functions at various stages of the build. Might be worth looking into.

I think that I'd also prefer a `AddMooringLine()` function instead of `SetNumberOfMooringLines(3)` and then specifying an index number. Then the `SetMooringLine*()` functions could just set the values in the last index of the vector (after checking that there is at least one mooring line). This will visually separate the mooring lines functions into groups and prevents a whole class of indexing issues.

By the way, the functions to set the fair lead velocity and acceleration need to be removed and some code added in `Build` to calculate those values based on the platform node velocity and acceleration, since they're rigidly connected. I'll have the necessary functions in the `model.hpp` class soon, but you can go ahead and drop those methods from the builder.

Thanks for putting this together.

src/interfaces/cfd/interface_builder.hpp
```diff
@@ -0,0 +1,130 @@
+#pragma once
+
+#include "floating_platform_input.hpp"
+#include "interface.hpp"
+#include "interface_input.hpp"
+#include "mooring_line_input.hpp"
+
+namespace openturbine::cfd {
+
+struct InterfaceBuilder {
```
faisal-bhuiyan, line 10: Should we call it `BuildInterface` to be consistent with how we named our other factory methods (such as `CreateBeams`)?

src/interfaces/cfd/floating_platform_input.hpp
```diff
@@ -21,7 +21,7 @@ struct FloatingPlatformInput {
     std::array<double, 6> acceleration{0., 0., 0., 0., 0., 0.};
 
     /// Platform point mass matrix
-    std::array<std::array<double, 6>, 6> mass_matrix;
+    std::array<std::array<double, 6>, 6> mass_matrix{};
 
     /// Mooring line array
     std::vector<MooringLineInput> mooring_lines;
```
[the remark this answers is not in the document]
ddement: I tried that.  Clang-Tidy had some reason for not liking it there.

src/interfaces/cfd/interface_builder.hpp
ddement: I was trying to be consistent with the Builder Pattern, where things are named this way by convention.  I'm open to changing it if there are strong opinions.

src/interfaces/cfd/interface_builder.hpp
```diff
@@ -0,0 +1,130 @@
...
+            acceleration;
+        return *this;
+    }
+
+    InterfaceBuilder& SetTurbine(const TurbineInput& turbine_in) {
+        interface_in.turbine = turbine_in;
+        return *this;
+    }
+
+    [[nodiscard]] Interface Build() const { return Interface(interface_in); }
```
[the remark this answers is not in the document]
ddement: Agreed.  I am thinking these guardrails will come once we have fully fleshed out the interface and added other features (such as an actual turbine).  I'd have to rely on yall to know what would make sense to check in this domain beyond certain obvious things (negative mass, etc).

src/interfaces/cfd/interface_builder.hpp
faisal-bhuiyan: No strong opinion here, wanted to make sure we are being consistent within the codebase.

ddement:
@deslaughter I've put together a prototype implementation of the a step-builder version of the interface.  I'd love your feedback on it.  See the FloatingPlatform regression test for a comparison of it versus the traditional builder.  Obviously, the implementation needs a little clean up, but I wanted to get your input before I blew away the old stuff.

deslaughter (approved):

PR #358, opened 2025-02-13
Fix the workflow error in main

faisal-bhuiyan:
Quick patch to fix the spack install workflow error currently in `main` branch.

### DIFF
--- a/.github/workflows/correctness-linux.yaml
+++ b/.github/workflows/correctness-linux.yaml
@@ -1,6 +1,9 @@
 name: OpenTurbine-CI
 
-on: push
+#on: push
+on:
+  push:
+    branches: [main]
 
 jobs:
   Correctness-Linux:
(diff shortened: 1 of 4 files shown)


faisal-bhuiyan:
@ddement Do you have anything in the pipeline for this fix? I was trying this, however looks like we need to directly push to `main` for this to work?

ddement:
> @ddement Do you have anything in the pipeline for this fix? I was trying this, however looks like we need to directly push to `main` for this to work?

@faisal-bhuiyan yes. I was planning to get to it tomorrow.  As you observed, we have to do it directly on the main branch, since spack has no visibility of our feature branches.

faisal-bhuiyan:
> @faisal-bhuiyan yes. I was planning to get to it tomorrow. As you observed, we have to do it directly on the main branch, since spack has no visibility of our feature branches.

Close this PR then?

PR #388, opened 2025-04-04
TPL linear solvers are now called directly rather than through Trilinos

ddement:
This PR introduces a major change to how sparse direct linear solvers are handled on OpenTurbine.  When configuring during CMake, you will need to specify a solver using the correct OpenTurbine_ENABLE_* flag, which will then load the requested TPL using find_package.

To use a linear solver, one must simply create the appropriate DSSHandle class and pass it along with the matrix/vector system into new free functions dss_symbolic, dss_numeric, and dss_solve.

Many options for linear solvers have been added: KLU, MKL, UMFPACK, and SUPERLU for CPU and cuSolverSP and cuDSS for CUDA.  More solvers will be added, including SUPERLU_MT for multithreading and rocSolver for rocm.


ddement:
Notes for Reviewers: 

Most of the real code changes are the addition of the linear solver code in `src/solver/linear_solver`.  Look at dss_handle.hpp, dss_symbolic.hpp, dss_numeric.hpp, and dss_solve.hpp to get an idea of the new API.  The other files are just the implementation details for each TPL - I recommend looking at KLU to understand what's been done.  All of the others follow the same format.  

solver/solver.hpp now has some compile-time logic for picking which solver to use.

Otherwise, the changes are just removing the need to use tpetra/amesos2 structures and calling our new functions.

deslaughter (approved):

PR #390, opened 2025-04-15
Remove linear node spacing option on blade interface

deslaughter:
The blade interface included an option to space the nodes linearly along the beam instead of at the GLL points. Based on feedback, this option is being removed and the StaticCurledBeam regression test results are updated to match (minor numerical differences)


ddement (approved):

deslaughter:
These changes are being incorporated in #387

PR #392, opened 2025-04-18
Changed all usages (where approriate) of create_mirror to create_mirror_view.

ddement:
On CPU-only runs, this should remove unnecessary allocations and copies.  It also avoids unnecessary initialization of views, which should provide (modest) improvements when running tests on both CPU and GPU.

Streamlined unit test matrix creation by adding a CreateView function, which copies the data from a std::array with layout left to the appropriate devices view. This reduces the amount of boiler plate in these tests significantly


faisal-bhuiyan (approved):
Looks good to me -- however it appears you need to rebase off of `main`.

PR #394, opened 2025-04-24
Change views and mirror_views to not initialize to zero

ddement:
This required a few changes to ensure that we weren't assuming these initialized values in setup process


faisal-bhuiyan (approved):
Looks good, what did we gain from our troubles here?

ddement:
> Looks good, what did we gain from our troubles here?

@faisal-bhuiyan 
The main thing this gives us is slightly faster startup times for larger problems, especially on GPU.  The improvement on CPU is more marginal.  
As a secondary benefit, I found a couple places where we were either unknowingly assuming an initial value of 0 or the algorithm was just incorrect (and accidentally giving us correct answers because a value of 0 worked out).  
This is an unfortunate situation where doing the "right" thing is really verbose and ugly.

PR #395, opened 2025-04-29
Generate openturbine visualization from NetCDF outputs

faisal-bhuiyan:
Several major changes were made in this work

- NetCDF-based outputs are used to generate visualization via a post-processing script
- To facilitate above, mesh connectivity of the model can be written to a yaml file
- Old visualization code were removed, thereby removing the dependency on `vtk`


src/interfaces/blade/blade_interface.hpp
```diff
@@ -124,6 +145,70 @@ class BladeInterface {
             node.UpdateMotion(this->host_state);
         }
     }
+
+    //------------------------------------------------------------------------------
+    // Write outputs
+    //------------------------------------------------------------------------------
+    void WriteOutputs() const {
```
deslaughter, line 152: Is there a way to do this that doesn't require allocating a vector for each node at runtime? Could the NetCDF writer take the `HostState` as an input and write all of the data to the file?

src/model/model.hpp
```diff
@@ -419,13 +424,147 @@ class Model {
         return {state, elements, constraints, solver};
     }
 
+    //--------------------------------------------------------------------------
+    // Outputs
+    //--------------------------------------------------------------------------
+
+    /// @brief Set up the NetCDF output writer
+    void SetOutputWriter(const std::string& output_file, size_t num_qps) {
```
deslaughter, line 432: The model class shouldn't deal with writing output as that's not part of the model. I could see the model class having a function to write the elements' node connectivity information to a YAML file for use in visualization.

src/model/model.hpp
```diff
@@ -419,13 +424,147 @@ class Model {
...
+        this->output_writer_ = std::make_unique<util::NodeStateWriter>(output_file, true, num_qps);
+    }
+
+    void WriteQPOutputsAtTimestep(State& state, Beams& beams, size_t timestep) const {
```
deslaughter, line 436: Let's remove the ability to write quadrature point data for visualization, I don't think we need to support it at this point.

src/viz/generate_vtk_output.py
```diff
@@ -0,0 +1,647 @@
...
+            indices_to_extract = range(self.num_nodes)
+        else:
+            # Validate that all requested indices are within range
+            for idx in node_indices:
+                if idx < 0 or idx >= self.num_nodes:
+                    raise ValueError(f"Node index {idx} out of range (max: {self.num_nodes-1})")
+            indices_to_extract = node_indices
+
+        # Read all data from NetCDF at once for the requested timestep
+        position = [
```
deslaughter, line 128: This looks like a lot of manual data manipulation which Python is really slow at doing. I suggest looking into the `xarray` package (https://docs.xarray.dev/en/stable/user-guide/io.html) which can load and work with NetCDF files natively.

src/viz/generate_vtk_output.py
```diff
@@ -0,0 +1,647 @@
...
+            mooring_file,
+            'lines',
+            mooring_connections,
+            mooring_nodes
+        )
+
+        print(f"Wrote floating platform visualization for timestep {timestep} to {output_dir}")
+
+
+    def _visualize_rotor(
```
deslaughter, line 479: I expected this library to be more generic, taking an element connectivity file and exporting vtk files for each element. This seems really fragile having it tied to specific models and visualization. Probably because you haven't created the connectivity file, but I think that should be done before merging this.

src/model/model.hpp
faisal-bhuiyan: Created an `Outputs` class to handle the responsibility of writing outputs (if requested) to a file and composed this to the `Model` class. Does that look good to you?

src/interfaces/blade/blade_interface.hpp
faisal-bhuiyan: This is just 7 vectors being allocated here, not for each node though. And the NetCDF interfaces expect a vector at some point I believe.

```
        std::vector<double> x(num_nodes_);
        std::vector<double> y(num_nodes_);
        std::vector<double> z(num_nodes_);
        std::vector<double> i(num_nodes_);
        std::vector<double> j(num_nodes_);
        std::vector<double> k(num_nodes_);
        std::vector<double> w(num_nodes_);
```

src/model/outputs.hpp
```diff
@@ -0,0 +1,115 @@
...
+
+        auto host_q = Kokkos::create_mirror_view(state.q);
+        auto host_v = Kokkos::create_mirror_view(state.v);
+        auto host_vd = Kokkos::create_mirror_view(state.vd);
+        auto host_x = Kokkos::create_mirror_view(state.x);
+
+        Kokkos::deep_copy(host_q, state.q);
+        Kokkos::deep_copy(host_v, state.v);
+        Kokkos::deep_copy(host_vd, state.vd);
+        Kokkos::deep_copy(host_x, state.x);
```
ddement, line 60: use `Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), ...)` instead of creating the mirror view and then copying.

src/model/outputs.hpp
```diff
@@ -0,0 +1,115 @@
...
+          location_(location) {}
+
+    [[nodiscard]] std::unique_ptr<util::NodeStateWriter>& GetOutputWriter() {
+        return this->output_writer_;
+    }
+
+    [[nodiscard]] OutputLocation GetLocation() const { return this->location_; }
+
+    /// @brief Write node state outputs to NetCDF file at specified timestep
+    void WriteNodeOutputsAtTimestep(State& state, size_t timestep) const {
```
ddement, line 39: This seems to be replicating code in blades interface.  I'm thinking you could have a single function here that takes a HostState.  This would mean that it is the calling code's job to get a version of their data in the HostState and Outputs is just in charge of writing it out to file. (also, in that case, model would not own an Outputs, but rather it would be owned at the interface level)

src/model/outputs.hpp
```diff
@@ -0,0 +1,115 @@
...
+#include <tuple>
+#include <vector>
+
+#include "elements/elements.hpp"
+#include "state/state.hpp"
+#include "utilities/netcdf/node_state_writer.hpp"
+
+namespace openturbine {
+
+class Outputs {
```
ddement, line 15: It might make sense for this to live in the interface folder rather than the model folder

docs/sphinx/walkthrough/compiling.rst
```diff
@@ -278,7 +278,7 @@ CMake from the command line or through a GUI such as ccmake.
 - ``OpenTurbine_ENABLE_TESTS`` builds OpenTurbine's test suite
 - ``OpenTurbine_ENABLE_UNITY_BUILD`` uses unity builds to potentially decrease
   compilation time
-- ``OpenTurbine_ENABLE_VTK`` builds OpenTurbine with VTK support for
+- ``OpenTurbine_WRITE_OUTPUTS`` builds OpenTurbine with VTK support for
```
deslaughter, line 281: I don't see a need for a feature flag for writing outputs. Let's remove this option.

src/interfaces/blade/blade_interface.hpp
deslaughter: Let's have the `NodeStateWriter` take the number of nodes as an argument to the constructor so it can preallocate and store these vectors. We know the system size by the time the output writer is being created in the interface. That should save the allocating/deallocating in a loop. For larger models these could be rather sizable.

src/model/outputs.hpp
deslaughter: I agree with @ddement, Outputs shouldn't be in the `model` directory since it isn't really a part of the model. If you don't want it in `interfaces`, it might make sense inside `state` since it is mostly a way to output the state. However, it should be outputting the contents of `HostState` which is part of the interfaces to avoid having the copy the data twice.

src/model/outputs.hpp
deslaughter: Use the `HostState` class which already contains this data.

src/model/mesh_connectivity.hpp
```diff
@@ -0,0 +1,199 @@
...
+            keys.push_back(id);
+        }
+        std::sort(keys.begin(), keys.end());
+
+        // Add each element in sorted order of id
+        for (const auto& id : keys) {
+            YAML::Node array_node;
+            std::for_each(map.at(id).begin(), map.at(id).end(), [&array_node](const auto& node_id) {
+                array_node.push_back(node_id);
+            });
```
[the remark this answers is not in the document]
faisal-bhuiyan: `std::back_inserter` is not compatible with `YAML::Node` -- I tried that before settling on this

src/model/outputs.hpp
faisal-bhuiyan: `Outputs` class moved under `openturbine::interfaces` namespace -- now we can write outputs only via an interface.

docs/sphinx/walkthrough/compiling.rst
faisal-bhuiyan: Removed this flag -- now we write outputs by default for the interface r-tests.

src/interfaces/blade/blade_interface.hpp
faisal-bhuiyan: I think I found a good compromise  -- rather than allocating the vectors at each function call, I am declaring them as class attributes and allocating them inside the c-tor. Does that sound like a reasonable solution here?

faisal-bhuiyan:
@ddement We're getting close to merging this PR, looks like you will need to modify the `spack` package for openturbine since we removed `vtk` requirements.

deslaughter (approved):

ddement:
@faisal-bhuiyan You'll need to remove the deleted file from `src/interfaces/CMakeLists.txt` to get the install test to get past its current error.

PR #407, opened 2025-05-13
Modified kernels to allow execution on host even when Kokkos is configured for CUDA

ddement:
This is the culmination of a series of PRs to decouple where we run from the default Kokkos configuration.  It modifies all of the execution policies at each kernel launch to be based on the desired execution space as well as all of our scratch views.  It adds two regression tests which run on host even when built for CUDA.


deslaughter (commented):
I can't claim to understand the template programming, nor would I expecte anyone who hasn't been using Kokkos for a significant amount of time to understand these changes. I expect the error messages from using the wrong type in one of these templates to be mostly incomprehensible. How hard will it be to add/modify the code after this change?

ddement:
> I can't claim to understand the template programming, nor would I expecte anyone who hasn't been using Kokkos for a significant amount of time to understand these changes. I expect the error messages from using the wrong type in one of these templates to be mostly incomprehensible. How hard will it be to add/modify the code after this change?

@deslaughter 
This should be pretty easy to modify - compared to before, you just have to include the extra `DeviceType` template argument when creating a view.  This can be any of a `Kokkos::MemorySpace`, `Kokkos::ExecutionSpace`, or `Kokkos::Device` (a combination of both) type, so it's pretty hard to misuse it from the top level, and the template has already been threaded through everything, so in practice it will just be saying `DeviceType` without worrying about any details.  The only changes in the regression tests were adding new tests - no code what changed in existing tests.  In unit tests, the only changes were adding the DefaultExecutionSpace argument at kernel creation.  

The error messages can be verbose at compile time if you forget to include the `DeviceType` in a view and try to copy it to one with that argument (or vice-versa), but it is a pretty straight forward message at the top to the effect of "incompatible view type at line xyz".  If you forget the `DeviceType` template argument for your kernel ranges, it will likely result in runtime errors of attempting to access an inaccessible memory space.  Those are the only two errors that I ran into when putting this code together, and it should be (fairly) obvious from context clues (i.e. blindly copying the code around you) what's necessary to avoid these problems in the first place.  

Note that these bugs will only be exposed when running in a GPU build with a host execution space (see the added regression tests).  If you stick to a CPU build, everything you did before will still work without change.  This may make prototyping algorithms easier/less verbose, just make sure to add in the extra arguments so it will work for all configurations.  When debugging in a GPU build, the only errors will be in the two regression tests specifically targeting non-default execution spaces, which should make the nature of the bug obvious.  

Also, by design, this change is only at the "engine" level - nothing touches the interfaces (either the API or the implementation code).  You can write/edit high level code without any knowledge of this change.

deslaughter (approved):

PR #417, opened 2025-06-05
Use external ROSCO and ADI rather than building them ourselves

ddement:


faisal-bhuiyan (approved):
Looks good from a quick check. We might need to revisit this: https://github.com/Exawind/openturbine/tree/main/recipes

PR #433, opened 2025-07-10
Changed interface tests to no longer write to disk by default.

ddement:
This cleans up the directory where regression tests are written and removes the performance hit (about 10%) from the iteration loop by default.

Writing output can be reenabled by changing the write_output option to true.


deslaughter (commented):
We should leave this enabled for at least one interface.

PR #482, opened 2025-08-28
Upgraded to use a few new C++20 features

ddement:
Uses std::span, ranges, and the new numbers header where appropriate.

### DIFF
--- a/src/constraints/constraints.hpp
+++ b/src/constraints/constraints.hpp
@@ -2,6 +2,8 @@
 
 #include <array>
 #include <numeric>
+#include <ranges>
+#include <span>
 #include <vector>
 
 #include <Kokkos_Core.hpp>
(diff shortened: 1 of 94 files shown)


faisal-bhuiyan:
> Uses std::span, ranges, and the new numbers header where appropriate.

Why was this refactor necessary i.e. what are we gaining from making this switch from `std::vector`?

ddement:
> > Uses std::span, ranges, and the new numbers header where appropriate.
> 
> Why was this refactor necessary i.e. what are we gaining from making this switch from `std::vector`?

When we used std::vector in our interfaces, we're requiring the calling code to use that container.  This has the implication that it must use heap allocation and not use any kind of custom container. The truth is that we don't care how the memory is allocated or handled in these locations. This change decouples the choice and lets the calling code do whatever it wants - we just need a contiguous block of memory and a length. In practice, you can see this in our tests, where we can now use std::array to allocate known-sized containers at compile time on the stack. 

I do regret the need to use, for example, `std::span<std::array<double, 7>`. I'd love to be able to make it more generic, but the "good" way to do so would need mdspan, which doesn't come until c++23. We could require a `std::span<double>` with an assumed layout, but I didn't want to make a change like that in a big commit like this where it might get overlooked.

The use of iota for the loops was a way to make our looping structure more modern, even when c++ iterators don't make sense. For the most part, these are just simple replacements, even when a more sophisticated pattern with ranges could be used. I expect the performance to be identical to what we had before, and it has the benefit of making it obvious if we do something tricky with the looping. We don't do this in our main kernel code because it doesn't work on CUDA devices. 

@faisal-bhuiyan does this answer your question?

PR #483, opened 2025-09-03
Adds transliteration of aero.rs from Derek's rust code

ddement:


src/math/quaternion_operations.hpp
```diff
@@ -180,6 +181,21 @@ KOKKOS_INLINE_FUNCTION void QuaternionInverse(
     }
 }
 
+/**
+ * @brief Computes the inverse of a quaternion
+ */
+inline std::array<double, 4> QuaternionInverse(const std::array<double, 4>& quaternion) {
+    const auto length = std::sqrt(
```
deslaughter, line 188: Might need a check here for length being a NaN.

src/interfaces/turbine/turbine_interface.hpp
```diff
@@ -56,6 +56,23 @@ class TurbineInterface {
     /// @brief Restores the previously saved state (in correction step)
     void RestoreState();
 
+    /// @brief Return a reference of the model owned by this interface
```
deslaughter, line 59: What do you think of adding the Aerodynamic bodies to the Turbine interface so we don't have to expose HostState and Model? They would need to be optional, but I think it would make sense considering how they'll be used.

deslaughter (approved):
Looks pretty good. I had a couple of comments, but nothing major

src/interfaces/turbine/turbine_interface.hpp
ddement: Done.  This was necessary to get the timing of setting/resetting the forces correct.

deslaughter (approved):
Looks great, thanks for putting it together!

PR #487, opened 2025-09-18
Replaced vector and quaternion operations on CPU with those from Eigen library

ddement:


faisal-bhuiyan (dismissed):
I am approving this -- just make sure we are good to add this dependency to Kynema

PR #509, opened 2025-10-28
Enhance netcdf writing to allow buffering and optional state component writing

faisal-bhuiyan:
Refactored NetCDF-based output writing utility of Kynema to improve overall design, functionality, and efficiency:

- Get a user-provided input for which `State` components to write (default behavior is to write all of them)
- Use a buffer to write several time steps worth of data at a time
- Option to close and re-open the file on-demand
- Refactored the interfaces to use a separate class to define the output configuration via `OutputsConfig` and `OutputsBuilder`

Some snapshots of writing performance improvements from above items. We can expect a **5-10X improvement** in the writing performance for big simulations by taking advantage of writing only the data we need + using buffering to write the data intermittently.

Output from `NodeStateWriterTest.StateWritingPerformance_SelectiveVsAllStates`:

```
438: Selective State Writing Performance (100 nodes, 1000 steps):
438:   All states (x,u,v,a,f): 278 ms
438:   Position only (x): 47 ms
438:   Speedup: 5.91x
```

Output from `NodeStateWriterTest.StateWritingPerformance_BufferedVsUnbuffered`:
```
439: Performance of NodeStateWriter with 100 nodes and 1000 steps: 
439:  unbuffered : 54 ms
439:  buffered   : 35 ms
439:  speedup    : 1.54x
```

### DIFF
--- a/tests/unit_tests/interfaces/test_outputs.cpp
+++ b/tests/unit_tests/interfaces/test_outputs.cpp
@@ -14,8 +14,11 @@ namespace kynema::tests {
 using kynema::interfaces::Outputs;
 
 TEST(OutputsTest, ConstructWithAndWithoutTimeSeriesFile) {
-    const std::string node_state_file = "node_states_outputs.nc";
-    const std::string time_series_file = "time_series_outputs.nc";
+    const ::testing::TestInfo* const test_info =
+        ::testing::UnitTest::GetInstance()->current_test_info();
+    const std::string test_name = std::string(test_info->test_case_name()) + "_" + test_info->name();
+    const std::string node_state_file = "node_states_outputs_" + test_name + ".nc";
+    const std::string time_series_file = "time_series_outputs_" + test_name + ".nc";
     std::filesystem::remove(node_state_file);
     std::filesystem::remove(time_series_file);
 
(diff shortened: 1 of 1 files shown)


deslaughter:
After our discussion with Michael, I'm concerned that the buffering will cause issues. If we have to open the file, write, and close the file every time step, does the buffering still make sense?

ddement:
I don't know that buffering will cause issues - we can just set it to 1 time step and/or flush on file close.  I agree that it doesn't make sense for the "open-write-close" workflow, but it should work just fine (and I don't think there will be any performance issues).

faisal-bhuiyan:
I also don't see how buffering will cause an issue -- it does flush before going out of scope. But it does not provide any performance benefits if we are just closing the file at every time step, which was the intention behind adding this feature.

src/utilities/netcdf/node_state_writer.hpp
```diff
@@ -21,20 +21,44 @@ namespace kynema::util {
...
+ * timestep and node index. The file structure uses an unlimited time dimension to allow
+ * for continuous writing of timesteps during simulation.
+ *
+ * @note The class includes buffering to improve write performance by batching multiple
+ * timesteps together before writing to disk.
  */
 class NodeStateWriter {
 public:
+    /// Default buffer size (number of timesteps to accumulate before auto-flush)
+    static constexpr size_t kDefaultBufferSize{10};
```
ddement: I'm thinking this should be a default of 1 (or 0?), so that the default is to use the minimum amount of memory

deslaughter: I agree, by default it should write every step. If users need better performance, they can increase it.

PR #514, opened 2025-11-07
Dynamic verification tests for kynema

faisal-bhuiyan:
Added a couple of clamped beam under tip loading tests to our `tests/verification/` suite to serve as time domain cases:

- A straight uniform beam with two sections
- IEA 15 MW turbine blade

As opposed to the static cases, these tests will involve comparing the displacement history for a node over a period of time. For example, following is the comparison for the straight beam response with BeamDyn.

image


faisal-bhuiyan:
@ddement I have a suspicion the rosco controller is not being found in the CI runs, however it is failing silently. I could reproduce the error locally by not loading rosco with spack (however cmake did not complain). I am pushing a change to confirm this by making rosco a required library when `Kynema_ENABLE_ROSCO_CONTROLLER=ON`.

michaelasprague (approved):
I think this looks fine to me, but curious if @ddement has comments

PR #517, opened 2025-11-13
make certain mass elements optional; check using diagonal

mbkuhn:
@deslaughter 

from the NREL 5MW case:
leaving the yaw bearing quantities as zero leads to invalid answers, and I think this is because the yaw bearing is still created as a mass element. Instead of creating it with an inertia matrix of 0's, this checks part of the matrix to determine whether there is nonzero mass in the element and then chooses to add or not add the mass element.


mbkuhn:
Derek has already addressed this in a different branch

PR #544, opened 2026-04-15
modify flush command to do nothing if file is closed

mbkuhn:
#537 introduced a Flush() command as part of the time series writer destructor. This causes an error at the conclusion of AMR-Wind-coupled runs, because it attempts to flush and sync when the file has already been closed (which is how it is set up on the AMR-Wind side). Following the existing paradigm in the netCDF interface of kynema, where Open() and Close() do nothing if the file is already in that state, this PR makes Flush() do nothing if the file is already closed.

### DIFF
--- a/src/utilities/netcdf/netcdf_file.hpp
+++ b/src/utilities/netcdf/netcdf_file.hpp
@@ -57,6 +57,11 @@ public:
      */
     void Open();
 
+    /**
+     * @brief Checks if the NetCDF file is open
+     */
+    bool IsOpen() const { return netcdf_id_ != -1; }
+
     //--------------------------------------------------------------------------
     // Setter/Write methods
     //--------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


jrood-nrel (dismissed):
Awesome.

mbkuhn:
This is the netCDF error that occurs without this change:
libc++abi: terminating due to uncaught exception of type std::runtime_error: Failed to sync NetCDF file: NetCDF: Not a valid ID

The problem is really with the sync operation, not with the flush. On top of that, the FlushBuffer() part of the Flush() command should already be inactive because, with the file already closed, the buffer has been purged (and there's a conditional related to the buffer size within FlushBuffer())

Just changed things to only limit the sync, confirmed that still fixes the problem with AMR-Wind.

jrood-nrel (approved):

PR #565, opened 2026-05-27
Upgrades kynema to use Kokkos version 5.

ddement:
Applied API changes to the SIMD loading functions and Kokkos-Kernels batched copy.  The HostMirror type is also replaced by host_mirror_type on Kokkos Views.  Extra protection was added to prevent kernel vector lengths of 0 (this state was impossible to reach previously, but it triggered newly added checks in Kokkos).

### DIFF
--- a/.github/workflows/correctness-linux.yaml
+++ b/.github/workflows/correctness-linux.yaml
@@ -41,7 +41,7 @@ jobs:
           spack install superlu
           spack install rosco@2.9.0
           spack install openfast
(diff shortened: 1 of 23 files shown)


ddement:
Note that this commit requires a rebuild of the dependency cache (since it uses the old Kokkos version).  When someone goes to review/merge this, you should do so - I'm holding off to prevent conflicts with other PRs in progress.

jrood-nrel (approved):

PR #570, opened 2026-07-17
add ability to set time step index externally

mbkuhn:
without this, a restarted kynema sgf+fmb simulation always resumes with a time step of 0, which is confusing

### DIFF
--- a/.github/workflows/correctness-linux.yaml
+++ b/.github/workflows/correctness-linux.yaml
@@ -28,7 +28,10 @@ jobs:
           path: |
             ${{ github.workspace }}/spack
             ${{ github.workspace }}/spack-packages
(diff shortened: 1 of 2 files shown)


PorterJustin:
This looks like you are setting the correct variable/it is consistent with this line: https://github.com/kynema/kynema-fmb/blob/e9b00b4ccadb99fcd88b4f7d543e0da473700e95/src/step/predict_next_state.hpp#L46

The use of `time_step` is confusing at a glance in the repo since it looks like most of the time it is the length of a time step, not an index. Your function is clearly named though, so that's not an issue for this PR.

PorterJustin (dismissed):
Looks good to me.

